PR #12, opened 2014-06-12
Update requirements.txt

WallE256:
Before, it was giving this error http://pastebin.com/hJZzZWYZ


accek:
I didn't hear about this from >9 people who recently installed SIO2 in front of me. Assuming this is no more the case.

PR #18, opened 2015-09-26
(SIO-1668) Quickfix broken report view when tests are not available

matrach:


accek:
You already fixed it using Gerrit. :)

PR #32, opened 2018-09-01
Update setup.py

Hacper:
Issue: https://jira.sio2project.mimuw.edu.pl/browse/SIO-2116


setup.py
```diff
@@ -35,6 +35,7 @@
         "Django>=1.9,<1.10",  # when upgrading, upgrade also django-two-factor-auth!
         "pytz>=2013b",
         "sqlalchemy",
+        "django-otp==0.3.4",
```
Mrowqa: I would add comment with explanation why this version and call to action to update it when updating certain other dependencies - like I did with Django and two-factor.

PR #34, opened 2018-09-23
Lock django-debug-toolbar version

LiquidPL:
>=1.10 requires Django 1.11


accek:
Please add an appropriate comment, like in #32

PR #35, opened 2018-09-25
Fix mixed translation of workers dashboard

michcioperz:
Haven't tested it yet, hopefully @LiquidPL can test it during the contest he's running this week


michcioperz:
It was verified by @LiquidPL on his production that it does not explode.

PR #39, opened 2019-04-11
Group time limits in config.yml

hugo213:
Allows problem authors to set time limits for *groups* in `config.yml`. Time limits for groups can be overriden by time limits for single tests.

### DIFF
--- a/oioioi/sinolpack/package.py
+++ b/oioioi/sinolpack/package.py
@@ -736,7 +736,7 @@ class SinolPackage(object):
 
         if created:
             instance.time_limit = self.time_limits.get(name,
-                    self.time_limits.get(re.search('^\d+', name).group(0), DEFAULT_TIME_LIMIT))
+                    self.time_limits.get(group, DEFAULT_TIME_LIMIT))
 
         memory_limit = self._get_memory_limit(created, name)
         if memory_limit:
--- a/oioioi/sinolpack/package.py
+++ b/oioioi/sinolpack/package.py
@@ -736,7 +736,7 @@ class SinolPackage(object):
 
         if created:
             instance.time_limit = self.time_limits.get(name,
-                    DEFAULT_TIME_LIMIT)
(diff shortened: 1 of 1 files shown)


michcioperz:
Weird, I swear I saw a closed ticket on JIRA for adding this feature before.

michcioperz:
Ah, never mind. I see now what you did there, a shorthand syntax for assigning a time limit to a whole group rather than each test separately.

hugo213:
We have this little feature in Staszic and yesterday I heard that people want it on OI too

michcioperz:
I think the regex is ugly yet I can't think of a cleaner way to get the group number.
I think it would be good to write a simple test (could be just another variation of Sumżyce tgz package) to verify that it works.

hugo213:
There was indeed a cleaner way. Group number is stored in `group` variable which I was unaware of.

accek:
That's a nice improvement, thanks!

I think this one deserves a test, though. Existing mechanism was tested indirectly here:
https://github.com/sio2project/oioioi/blob/master/oioioi/sinolpack/tests.py#L291
but it may be easier to make a new one similar to this:
https://github.com/sio2project/oioioi/blob/master/oioioi/sinolpack/tests.py#L155
rather than changing the first one. What do you think?

hugo213:
To write any bigger piece of code I'd have to set up a dev server with your version of sio2. When I find enough free time to do so, I will submit the tests. I'm sorry, but because of many other responsibilites it might take me some time.

accek:
No hurry. :-)

If you just want to run tests, then setting up virtualenv and using the bundled requirements.txt should be enough.

michcioperz:
I think this might be a subset of https://github.com/sio2project/oioioi/commit/af30a2b4a551987803a92acecbbcc172c4d7ae66

DietPawel:
As this thread had no activity for past 2 years and similar improvements were implemented I am closing the pull request.
But fell free to reopen if the functionality is not there.

PR #71, opened 2022-05-27
(SIO-2489) End-to-end testing with Cypress

tudny:
End-to-end testing with Cypress framework. 

Added commands to `easy_toolbox.py` and new script file `test_cypress.sh` to run tests. 
For now, it is required to run tests on clean instance of OIOIOI with flushed database and one superuser (admin, admin).

Added Github Action workflow which runs on every Monday, Wednesday and Friday. It requires about 15 minutes to build the image with `docker-compose` (using `easy_toolbox.py`). [Idea: add built image to Docker Hub and workflow, doing it once in a while]

Future upgrade would include a new way to run test server with separate database (e.g. SQLite like in `test.sh`).

Thanks to Bartłomiej Chmiel for the whole idea and PoC.


easy_toolbox.py
```diff
@@ -28,20 +28,20 @@
     ("build", "Build whole OIOIOI from source.", "build", True),
     ("up", "Run all SIO2 containers", "up -d"),
     ("down", "Stop all SIO2 containers", "down", True),
-    ("run", "Run server", "exec web python3 manage.py runserver 0.0.0.0:8000"),
-    ("bash", "Open command prompt on web container.", "exec web bash"),
-    ("bash_db", "Open command prompt on database container.", "exec db bash"),
+    ("run", "Run server", "exec{} web python3 manage.py runserver 0.0.0.0:8000"),
```
twalen, line 31: this is quite cryptic, maybe it might be easier to use named params and to have {exec} param:
```"{exec} web python3 manage.py runserver 0.0.0.0:8000"```

and `{exec}` = `"exec"` or `"exec -T"` depending on the context

File: oioioi_cypress/cypress/integration/contest/settings.spec.ts [hunk 2]
```diff
@@ -0,0 +1,71 @@
...
+  });
+});
+
+const loginAsAdmin = () => {
+  cy.visit("/");
+  cy.get(".username").click();
+
+  cy.fixture("credentials").then((data) => {
+    cy.get("#navbar-login-input").type(data.user.username);
+    cy.get(":nth-child(4) > .form-control").type(data.user.password);
```
twalen, line 44: is it possible to access this element by name like:
`.get("input[name='auth-password']")`

oioioi_cypress/cypress/integration/user/submit.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  it("should submit task", () => {
+    addProblem();
+    submit();
+    validateIfStatusOk();
+  });
+});
+
+const loginAsAdmin = () => {
+  cy.visit("/");
```
twalen, line 32: actually why not simply use: `cy.visit("/login/")`

oioioi_cypress/cypress/integration/user/submit.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+  cy.get(".username").click();
+
+  cy.fixture("credentials").then((data) => {
+    cy.get("#navbar-login-input").type(data.user.username);
+    cy.get(":nth-child(4) > .form-control").type(data.user.password);
```
twalen, line 37: the same as before selecting by name might be simpler

oioioi_cypress/cypress/integration/contest/settings.spec.ts
twalen: actually why not use `cy.login(data.user)`

oioioi_cypress/cypress/integration/user/submit.spec.ts
twalen: `cy.login(data.user)`

test_cypress.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
```
twalen, line 2: in README we need to mention that yarn + npm is required to run Cypress.

twalen (approved):
Add 2 small comments, that can be addressed in near future (but not blocking merge).

PR #87, opened 2022-08-13
(no-ticket) easy_toolbox.py added to README

tudny:
Added a note about `easy_toolbox.py` to README, as it may not be obvious to use it during development.

Added a note about INSTALL being deprecated method of installation - it does not work for some time.


twalen (commented):
looks good

twalen (approved):

PR #92, opened 2022-10-04
(SIO-2491) No avatar for admin created users

tudny:
Testing evicence
image


Michcioperz:
???
Considering that the template code you're editing is guarded behind `{% if avatar %}`, isn't this a backend problem instead? https://github.com/sio2project/oioioi/blob/6af8ef04dfff543f711a8b62dc424b5ee56447a9/oioioi/base/processors.py#L36-L44

oioioi/base/templates/ingredients/navbar-user.html
```diff
@@ -1,4 +1,4 @@
-{% load i18n menu get_su get_user_info_form get_maintenance_state %}
+{% load i18n menu get_su get_user_info_form get_maintenance_state static %}
```
Michcioperz, line 1: This change is now no longer necessary too

twalen (approved):

tudny:
I'm not sure if I like the result. When we do it in backend the icon looks like nobody is logged in. If email is null we may send something like "example@example.com" and get expected result, but it looks ugly in the code. I think somebody should make a design decision. 

a) send mock email if null -> get expected result (ugly) code 
b) leave as is (d55a733) -> icon look same as when user is not logged in 

@twalen @Michcioperz 
What do you think?

oioioi/base/processors.py
```diff
@@ -34,7 +34,7 @@ def mathjax_location(request):
 
 
 def gravatar(request):
-    if request.user.is_authenticated:
+    if request.user.is_authenticated and request.user.email:
 
         def generator():
             return str(get_gravatar_url(request.user.email, size=25)) or ''
```
Michcioperz, line 40: > a) send mock email if null -> get expected result (ugly) code

I was about to suggest this:
```suggestion
            return str(get_gravatar_url(request.user.email or '', size=25))
```

but it seems that Gravatar then responds with 404 (purely out of spite? Only happens with .jpg suffix which django-gravatar seems to add. Like, it's fine with an all-zeros hash, but not with hash of empty string). So I guess you have to have *some* default still.

Anyway since Gravatar only receives hashes you should be able to use a non-email-like string such as "oioioi"? I don't really see that as ugly, and certainly it sounds more predictable than using an email.

(Please verify what I'm saying before proceeding, cause I don't have a dev env set up, I'm just pasting chunks of Python into REPL and looking to see what happens.)

tudny: Yes, you are right. I checked it with 'oioioi' and Gravatar produces expected image as a result. Thanks for the help! 
Pushed and awaites @twalen's review.

twalen (approved):

PR #96, opened 2022-10-17
(no-ticket) Toolbox feature - inquirer only for GUI 

tudny:
inquirer is not required when using `./easy_toolbox.py <arg>` 
The package is imported only when using ~~CLI~~ GUI

### DIFF
--- a/easy_toolbox.py
+++ b/easy_toolbox.py
@@ -2,7 +2,7 @@
 
 # pip requirements:
 #   python ^3.6
-#   inquirer
+#   inquirer     (only for CLI)
 # 
 # system:
 #   docker
@@ -17,7 +17,6 @@
 
 import sys
 import os
-import inquirer
 
 
 BASE_DOCKER_COMMAND = "OIOIOI_UID=$(id -u) docker-compose" + \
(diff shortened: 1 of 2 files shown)


tudny:
Also, added `.fleet` to `.gitignore` as someone may use it. Same as adding `.vscode` some time ago.

easy_toolbox.py
```diff
@@ -110,6 +109,7 @@ def get_action_from_args() -> Option:
 
 
 def get_action_from_cli() -> Option:
```
twalen, line 111: actually this is a good moment to fix the function name: `get_action_from_cli` -> `get_action_from_gui`

easy_toolbox.py
```diff
@@ -2,7 +2,7 @@
 
 # pip requirements:
 #   python ^3.6
-#   inquirer
+#   inquirer     (only for CLI)
```
twalen, line 5: CLI (Command Line Interface) -> GUI

tudny: done

easy_toolbox.py
tudny: done

twalen (approved):

Issue #102, opened 2022-11-20
Button to copy submission's code to clipboard is not translated after being clicked.


sommbre:
Button to copy submission's source code is not translated after being clicked.

How to reproduce:
1. Set system language to polish.
2. Go to any submission.
3. Click 'Pokaż kod'.
4. On new page click 'Skopiuj do schowka'.
5. The button's text is now in English ('Copied').

The problem is present on fresh oioioi instance and on szkopul too.

The text should be translated to give a more polished user experience.

![image](https://user-images.githubusercontent.com/55536129/202894755-d5dbce6f-8244-40df-bb20-07328e09b429.png)
![image](https://user-images.githubusercontent.com/55536129/202894776-033476b6-83c6-46ff-a670-886138875a5e.png)

PR #111, opened 2022-11-27
Fix for issue #102

sommbre:
Fix for #102. I ran `django-admin makemessages` locally and added translation (included only my change) but didn't compile them to `*.mo` file so we need to run Transifex jobs.

Or I can do it locally if we don't want to run these jobs for some reason

### DIFF
--- a/oioioi/programs/static/common/clipboard.js
+++ b/oioioi/programs/static/common/clipboard.js
@@ -3,7 +3,7 @@ function copyCodeToClipboard() {
         const button = document.getElementById("cpy_btn");
         button.classList.remove("btn-outline-secondary");
         button.classList.add("btn-success");
-        button.textContent = gettext("Copied!");
+        button.textContent = "Copied!";
     }, function() {
         alert("Unable to copy Code");
     });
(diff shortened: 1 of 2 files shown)


oioioi/programs/static/common/clipboard.js
```diff
@@ -3,7 +3,7 @@ function copyCodeToClipboard() {
         const button = document.getElementById("cpy_btn");
         button.classList.remove("btn-outline-secondary");
         button.classList.add("btn-success");
-        button.textContent = "Copied!";
+        button.textContent = gettext("Copied!");
     }, function() {
         alert("Unable to copy Code");
```
tudny, line 8: How about fixing this miss translation as well.

tudny, line 8: How about fixing this missing translation as well?
line 8

tudny (dismissed):

tudny:
Translations are done by the GH Action. It should be run manually, but it calls `makemessages`. All translation changes should be done using workflows, as it keep the consistency, such as ignoring unused strings. 

The mentiond workflow:
https://github.com/sio2project/oioioi/blob/de6026fbb889a37ff07866dbab60d87c5ee941ed/.github/workflows/translations-upload.yml#L60

twalen (approved):

PR #115, opened 2022-12-12
Move the sum to the left in the ranking page

otargowski:
This will increase readability, as the users won't have to fling their eyes between the edges of the screen to read the sums of points.
For contests with a lot of problems, this will make an even bigger difference, because you won't have to scroll horizontally.


twalen:
I'm in favour of current UI I recommend to create mockups for old/new ranking and test it with end-users.

A-dead-pixel:
I created the mockups, both from real and anonymized data.
One is from the last Algorythmic Engagements and the other one shows a contest with a lot of problems.
They can be seen or downloaded [here](https://otsrv.net/stuff/sioranking/). 

How do you think the test should be conducted?

A-dead-pixel:
One can view the changed version also at https://szkopul.edu.pl/c/mistrz-programowania-2023/ranking/, since this has already been done in Mistrz Programowania:
https://github.com/sio2project/oioioi/blob/7118806a11ff39ec90f80f3b0607f3448b8e6db5/oioioi/mp/controllers.py#L174-L176

MasloMaslane:
I created surveys on discord community servers and most people (91/97) would prefer the sum on the left. It would be nice to have an option for the user to move the sum, but with how rankingsd works I don’t see a nice way of doing this

twalen:
> I created surveys on discord community servers and most people (91/97) would prefer the sum on the left. It would be nice to have an option for the user to move the sum, but with how rankingsd works I don’t see a nice way of doing this

vox populi, vox Dei 

let's merge this change.

twalen (dismissed):

A-dead-pixel:
Updates to the PR:
1. Logic for moving the sum to the left in Mistrz Programowania has been removed.
2. The background of the sum column has been made gray to improve the contrast, as per @MasloMaslane's request.

MasloMaslane (approved):
You can squash merge after the tests pass

PR #125, opened 2022-12-27
Better displaying countdown text

### LINKED ISSUE #122: Navbar round name visibility should be prioritized over time full format visibility and shouldn't be clipped when free space is available
1)
obraz
should be
obraz

2)
obraz
should be
obraz

KajetanLewandowski:
Closes #122


twalen (approved):
OK

PR #128, opened 2023-01-06
Introduce contest controller for Mistrz Programowania

geoff128:


oioioi/default_settings.py
```diff
@@ -217,6 +217,7 @@
     'oioioi.workers',
     'oioioi.quizzes',
     'oioioi._locale',
+    'oioioi.mp',
```
DietPawel, line 220: I am not sure all oioioi instances should have this module enabled by default. I would follow the convention with simillar `oioioi.pa` module which does not appear on this list.

oioioi/mp/controllers.py
```diff
@@ -0,0 +1,198 @@
...
+    show_email_in_participants_data = True
+
+    def registration_controller(self):
+        return MPRegistrationController(self.contest)
+
+    def ranking_controller(self):
+        return MPRankingController(self.contest)
+
+    def update_user_result_for_problem(self, result):
+        """
```
DietPawel, line 112: Sad empty docstring :/

oioioi/mp/models.py
```diff
@@ -0,0 +1,32 @@
+# coding: utf-8
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+from oioioi.base.utils.deps import check_django_app_dependencies
+
+from oioioi.participants.models import RegistrationModel
+from oioioi.contests.models import Contest
+
+check_django_app_dependencies(__name__, ['oioioi.participants'])
```
DietPawel, line 10: nice we check for `oioioi.participants`, but why not checking for `oioioi.contests`

oioioi/mp/models.py
```diff
@@ -0,0 +1,32 @@
...
+
+class MPRegistration(RegistrationModel):
+    terms_accepted = models.BooleanField(_("terms accepted"), default=False)
+
+    def erase_data(self):
+        self.terms_accepted = False
+        self.save()
+
+
+class SubmissionScoreMultiplier(models.Model):
```
DietPawel, line 21: This is a very good basic model. Out of scope for MP this model could be much more universal. Allowing for both start and end dates. Linking multipliers to rounds or even ProblemInstances would make it more versatile too.

oioioi/mp/urls.py
```diff
@@ -0,0 +1,9 @@
+from django.urls import re_path
+
+from oioioi.mp import views
+
+app_name = 'mp'
+
+contest_patterns = [
+    #re_path(r'^contest_info/$', views.contest_info_view, name='contest_info')
```
DietPawel, line 8: Not sure this line or this file is necessary at all.

oioioi/mp/templates/mp/ranking.html
```diff
@@ -0,0 +1,53 @@
+{% load i18n pagination_tags get_user_name static simple_filters %}
```
DietPawel, line 1: This file looks like 1:1 to `oioioi/oioioi/rankings/templates/rankings/default_ranking.html `. If it is unmodified I think the code should be restructured not to repeat ourself's

oioioi/mp/templates/mp/registration-notice.html
```diff
@@ -0,0 +1,8 @@
+{% load i18n %}
```
DietPawel, line 1: I know PA and OI copied this notice, but third time in a single codebase seems to indicate a structure issue. It might be fine for now, but an issue should be raised at least.

oioioi/mp/templates/mp/registration.html
```diff
@@ -0,0 +1,40 @@
+{% extends "base-with-menu.html" %}
```
DietPawel, line 1: If we depend on `oioioi.participant` we should be able to use its templates. This is again 1:1 copied file from oioioi/oioioi/participants/templates/participants/registration.html

oioioi/mp/models.py
```diff
@@ -0,0 +1,32 @@
...
+    even after round ends, until end_date
+    
+    Result score for submission after round's end is multiplied by
+    given multiplier value
+    """
+    contest = models.OneToOneField(
+        Contest, verbose_name=_("contest"), on_delete=models.CASCADE
+    )
+    multiplier = models.FloatField(_("multiplier"))
+    end_date = models.DateTimeField(_("end date"))
```
DietPawel, line 32: A newline at the end will not hurt anybody :)

oioioi/mp/controllers.py
```diff
@@ -0,0 +1,198 @@
...
+        return not MPRegistration.objects.filter(
+            participant__contest=request.contest
+        ).exists()
+
+
+class MPContestController(ProgrammingContestController):
+    description = _("Mistrz Programowania")
+    create_forum = False
+
+    show_email_in_participants_data = True
```
DietPawel, line 103: I will have to make sure we can do that from legal point of view.

oioioi/mp/controllers.py
```diff
@@ -0,0 +1,198 @@
...
+
+        if submissions:
+            best_submission = None
+            for submission in submissions:
+                ssm = SubmissionScoreMultiplier.objects.filter(
+                    contest=submission.problem_instance.contest,
+                )
+                
+                score = FloatScore(submission.score.value)
+                rtimes = self.get_round_times(None, submission.problem_instance.round)
```
DietPawel, line 129: I understood this logic, but it is not trivial and should have some helpful comments added at least. Perhaps it could be turned into separate function to make this method more compact, like apply_submissionscoremultiplier

oioioi/mp/controllers.py
```diff
@@ -0,0 +1,198 @@
...
+            TermsAcceptedPhraseAdminMixin,
+        )
+
+    def can_change_terms_accepted_phrase(self, request):
+        return not MPRegistration.objects.filter(
+            participant__contest=request.contest
+        ).exists()
+
+
+class MPContestController(ProgrammingContestController):
```
DietPawel, line 99: This whole class is quite big as a result of ProgrammingContestController being quite low-level. Overall it might be fine for our purposes

oioioi/mp/controllers.py
```diff
@@ -0,0 +1,198 @@
...
+from oioioi.participants.controllers import ParticipantsController
+from oioioi.participants.models import Participant
+from oioioi.participants.utils import is_participant
+from oioioi.programs.controllers import ProgrammingContestController
+from oioioi.rankings.controllers import DefaultRankingController
+
+CONTEST_RANKING_KEY = 'c'
+
+
+class MPRegistrationController(ParticipantsController):
```
DietPawel, line 21: I feel that some methods here are overridden just like in other similar apps. Should we restructure the code to use some Mixins for avoiding to repeat same code over and over again?

DietPawel (commented):
I think this code is ok. I would run without any issues. Most of my comments are about repeating code present in other apps, which is result of a codebase structure. The rest are minor errors/suggestions.

I would like to hear from @twalen, as we need this changes quickly in production.

oioioi/mp/controllers.py
DietPawel: It turned out that the organizers already signed data processing agreements and can access participants personal data.

DietPawel (dismissed):
Let us see where it goes...

oioioi/mp/admin.py
```diff
@@ -36,20 +36,24 @@ def get_actions(self, request):
...
 
 
 class SubmissionScoreMultiplierAdminMixin(object):
-    """Adds :class:`~oioioi.mp.SubmissionScoreMultiplier` fields to an
-    admin panel.
+    """Adds :class:`~oioioi.mp.models.SubmissionScoreMultiplier` to an admin panel
+    when contest controller is MPContestController.
     """
 
     def __init__(self, *args, **kwargs):
```
DietPawel, line 50: That's unneeded

oioioi/mp/admin.py
```diff
@@ -36,20 +36,24 @@ def get_actions(self, request):
...
         super(SubmissionScoreMultiplierAdminMixin, self).__init__(*args, **kwargs)
-        self.inlines = self.inlines + [SubmissionScoreMultiplierInline]
+
+    def get_inlines(self, request, obj=None):
```
DietPawel, line 53: Is it safe to run this function multiple times? I am not sure django guaranties it to be executed only once. Maybe just return self.inlines with SCMInline appended.

oioioi/mp/controllers.py
```diff
@@ -97,7 +97,7 @@ def can_change_terms_accepted_phrase(self, request):
 
 
 class MPContestController(ProgrammingContestController):
-    description = _("Mistrz Programowania")
+    description = _("Master of Programming")
```
DietPawel, line 100: Not sure this name should be translated to English. It is more PR related question rather than an issue.

oioioi/mp/templates/mp/registration.html
```diff
@@ -11,7 +11,7 @@
     {{ form.media.js }}
 {% endblock %}
 
-{% block title %}{% trans "Register to the contest" %}{{ contest_name }}{% endblock %}
+{% block title %}{% trans "Register to the contest" %} {{ contest_name }}{% endblock %}
```
DietPawel, line 14: Is this result of running static analysis (from the README)?

DietPawel (approved):

Issue #135, opened 2023-01-15
Improve look of the source code preview


Aleshkev:
I propose the source code preview be changed to a cleaner look.

_Current_
![image](https://user-images.githubusercontent.com/43296251/212565443-7c4d3410-6577-47c1-8e01-eddcf71880e3.png)

_Proposed_
![image](https://user-images.githubusercontent.com/43296251/212565122-c1251a2e-1111-4848-8bbd-7d3890a4e6b9.png)

Additionaly, right now line numbers can be selected, however as they are in a separate table cell, this is nearly never useful. It also doesn't follow the behaviour of most tools for viewing source code which don't allow selecting line numbers at all.

_Current_
![c17841d6-f0e8-4ece-a070-62f3d96275a7](https://user-images.githubusercontent.com/43296251/212565514-669bfbf9-8625-4d8c-b83d-c4864b601864.gif)

_Proposed_
![3e95c903-e428-4f16-861e-966d63d753b1](https://user-images.githubusercontent.com/43296251/212565273-c0987192-ef2e-431a-a22b-3fc08fda8664.gif)

PR #136, opened 2023-01-15
Improve look of the source code preview

Aleshkev:
Resolves https://github.com/sio2project/oioioi/issues/135.

### DIFF
--- a/oioioi/programs/static/common/programs.scss
+++ b/oioioi/programs/static/common/programs.scss
@@ -33,6 +33,37 @@
 }
 
 .syntax-highlighttable {
+  // Collapse the border between line numbers and code.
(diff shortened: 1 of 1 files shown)


oioioi/programs/static/common/programs.scss
```diff
@@ -33,6 +33,37 @@
...
+  }
+  td.code {
+    & > .syntax-highlight {
+      border-top-left-radius: 0;
+      border-bottom-left-radius: 0;
+    }
+    padding-left: 0
+  }
+
+  // Disallow selecting line numbers. We disable pointer events so that a mouse
```
twalen, line 54: nice catch

twalen (approved):

Issue #84, opened 2022-06-28
Last submissions go beyond page


tudny:
Last submissions pop-up goes beyond the visible page when opened for the first time.

![image](https://user-images.githubusercontent.com/17858116/176311854-e02ad661-6e17-439f-ac9a-ae3817d8c2d9.png)
![image](https://user-images.githubusercontent.com/17858116/176311956-1d748e67-bb66-4ab2-b65f-a36ad9bafa03.png)

tudny:
@mgr0dzicki looks like your Bootstrap aftermath

PR #145, opened 2023-01-31
Automatic submission(s) page refreshing on status update

### LINKED ISSUE #104: Add automatic submission page refresh
It would be nice if the submissions list/submission page would automatically refresh when the submission status changes.

etiaro:
Resolves #104 
Musiałem minimalnie dostosować testy, bo dodanie nowej funkcji "status_registry" czasem zwiększa liczbę zapytań

### DIFF
--- a/oioioi/base/tests/tests.py
+++ b/oioioi/base/tests/tests.py
@@ -100,7 +100,7 @@ class TestIndex(TestCase):
             response = self.client.get('/', follow=True)
         self.assertNotContains(response, 'test_user')
         self.assertTrue(self.client.login(username='test_user'))
-        with self.assertNumQueriesLessThan(70):
+        with self.assertNumQueriesLessThan(71):
             response = self.client.get('/', follow=True)
         self.assertContains(response, 'test_user')
         login_url = reverse('login')
@@ -135,7 +135,7 @@ class TestIndex(TestCase):
             response = self.client.get('/', follow=True)
             self.assertNotContains(response, 'navbar-login')
             self.assertNotContains(response, 'System Administration')
-        with self.assertNumQueriesLessThan(85):
+        with self.assertNumQueriesLessThan(87):
             self.assertTrue(self.client.login(username='test_admin'))
             response = self.client.get('/', follow=True)
             self.assertNotContains(response, 'navbar-login')
--- a/oioioi/contests/static/contests/submission_update.js
+++ b/oioioi/contests/static/contests/submission_update.js
@@ -1,0 +1,17 @@
+$(function() {
+  var contest_submissions = {};
+  $(window).on('updateStatus', function(event, data) {
+    if (data.contest_submissions) {
+      if (data.contest_submissions.length != contest_submissions.length || 
+          data.contest_submissions.some((el, i) => el.status != contest_submissions[i].status))
+        location.reload();
+          
+      contest_submissions = data.contest_submissions;
+    }
+  });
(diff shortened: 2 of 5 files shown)


A-dead-pixel:
Isn't the status fetched only every 5 minutes for non-superusers?
I think this would be much more useful if it utilised data from the notifications module.

etiaro:
> Isn't the status fetched only every 5 minutes for non-superusers? I think this would be much more useful if it utilised data from the notifications module.

Right, I forgot to check how often it happens for standard users. I'll look into this notifications module

etiaro:
> Isn't the status fetched only every 5 minutes for non-superusers? I think this would be much more useful if it utilised data from the notifications module.

Do you know how to get notifications working in local environment?
Or even how to get them working on szkopuł as regular user?

A-dead-pixel:
I will write the deployment instructions later today.
As for szkopuł, they are always on if you aren'y viewing from your phone. They are the thing showed in #84. You access them by clicking the yellowish dropdown button on the navbar.

etiaro:
Okay, I got it running, but the notifications about initial judge and finished judging are not working (NotificationsHandler is not called on these logs)

etiaro:
Got notifications working in really unconvienient way (set up and run rabbitmq and notifications_server in web container's bash), and they work alright, automatic refresh can be easily implemented. 
I'm not sure if I should also refresh scores list page?
Also, tried to replicate everything in Szkopuł, on never used account and browser, and it didn't work. Notification server is running and connecting, but the only incoming notification is "ERR_AUTH_FAILED" after a few seconds

A-dead-pixel:
Ideally, the new submission/result would be inserted into the list, otherwise refreshing the page should be decent. I'm not sure whether the other submissions for given problem below should refresh the page in the second case.

I checked, and it seems you're right: notifications on szkopuł are broken. Someone responsible should be informed.

The settings I use to enable notifications for local deployment:
- expose port 7887 in docker-compose
- settings.py:
```
NOTIFICATIONS_SERVER_ENABLED = True
NOTIFICATIONS_RABBITMQ_URL = 'amqp://oioioi:oioioi@broker'
NOTIFICATIONS_SERVER_URL = 'http://localhost:7887/'
```

Other defaults are fine I think. You don't have to run another rabbitmq instance.
This isn't exactly needed, but I also install nodejs and npm in the dockerfile, as opposed to leaving the installation to `notifications_server`.

etiaro:
Thanks! I'll go with implementing the automatic updates at the weekend.
I think it would be useful to add these instructions to notifications page at https://oioioi.readthedocs.io or at least change the default values in settings.py template, but I have no idea how or where can I do that

A-dead-pixel:
I have no experience with readthedocs, but I think the notifications/README.rst should be updated instead, at least to mention the port exposure and to clear up what settings need to be changed. It seems that the readthedocs stuff is more directed at APIs and the inner workings of oioioi. Nevertheless, it wouldn't hurt to mention the README there.

I don't know where the default settings for the rabbitmq connection should be changed, but it would be safest to add it to the sed command in the Dockerfile, where other docker-specific configuration changes reside, though it isn't too consistent with the deprecation of non-docker deployments.
That aside, could you convert this PR to a draft for now?

etiaro:
Sure, tbh I'm gonna make another one because these changes are going to trash

etiaro:
Moved to #146

PR #153, opened 2023-03-07
Fixes in MPContestController

geoff128:
1. There was an error when participant tried to access contest with problem with no round
2. Scores were not visible on problems list


A-dead-pixel:
It would be nice if the ranking had the usual colors. I think they are missing because of the non-integer scores, which in my humble opinion are not needed.

oioioi/mp/controllers.py
```diff
@@ -139,8 +139,20 @@ def update_user_result_for_problem(self, result):
...
                     best_submission = [submission, score]
 
+            try:
+                report = SubmissionReport.objects.get(
+                    submission=best_submission[0], status='ACTIVE', kind='NORMAL'
+                )
+            except SubmissionReport.DoesNotExist:
+                report = None
+
             result.score = best_submission[1]
```
twalen, line 149: I suggest to change typo of best_submission from tuple to normal variable and use following logic
- best_submission (currently best submission)
- add new variable best_submission_score (initially None, currently best score)
This would make this part of the code much more readable.

twalen (changes requested):
The changes looks good, but could you improve a little bit readability of `update_user_result_for_problem`?

DietPawel:
@geoff128 Could You please try to look into this by EOY?

MasloMaslane (approved):

twalen (approved):

PR #134, opened 2023-01-15
Refactor docker setup

LiquidPL:
Resolves #126.

This PR cleans up all the Docker and Docker Compose files used by oioioi.

Notable changes:
* reworked `Dockerfile` to use multi-stage builds, now the resulting image doesn't have any leftovers from the build process,
* the app is now running inside a virtualenv,
* it is now possible to build a smaller, production-ready image that doesn't include `apt` installed compilers,
* removed sandboxes from the development image, as mounting a volume/directory to the already existing filetracker database in the image would lead to confusing behaviour (and additionally, the sandboxes would be pointlessly duplicated in every image),
* added instructions on how to install the sandboxes on a machine without internet connection
* various settings (database, rabbitmq, judging) that were previously hardcoded are now configurable via environment variables (with original hardcoded values being now configured in compose files)
* removed duplicated docker container init scripts
* moved all scripts that are not being run on the host machine to a separate `scripts/` directory
* removed duplicated configuration in `docker-compose-selenium.yml`

Some additional things to consider:
* Is the selenium test suite actively maintained? I've made sure that it at least starts up properly, but from what I've seen all the tests fail and the suite hasn't been touched in a few years.
* Why is the development server being started as a separate command instead of being managed in supervisor? I haven't seen anything that would suggest it can't be run that way, and this way we could simplify the docker-compose commands by removing `extra/sample_configs/basic_settings_noserver.py`.
* This PR still lacks scripts and GitHub Actions workflows to build and publish the Docker images, as I'd prefer to discuss where to publish those images beforehand.
* There's also a PR for sioworkers at sio2project/sioworkers#12.

PR #161, opened 2023-03-16
Settings refactor

### LINKED ISSUE #160: Settings refactor
Simplify settings files in repository.

Suggested workflow:
- [x] remove basic_settings.py, all related files and configuration
- [x] Update comments/documentation of all settings in settings.py/default_settings.py
- [x] Move part of settings.py.template into default.settings.py (keep in mind to not break tests, or anything else that imports default_settings)
- [x] use environment variables to override default values in default_settings.py
- [x] make generation of settings.py(settings.py.template) a copy of default_settings.py with commented out values
- [x] look for possible secrets and load them from environment variables
- [ ] create environment variables table somewhere in documentation
- [ ] update docker config to use some .env file
- [ ] analyze possible migration problems after change of default settings.
- [ ] create script for settings.py automatic modification, overriding default_settings.py values with old ones based on CONFIG_VERSION in settings.py

etiaro:
Resolves #160


Dockerfile
```diff
@@ -68,8 +68,7 @@ RUN oioioi-create-config /sio2/deployment
 WORKDIR /sio2/deployment
 
 RUN sed -i -e \
```
LiquidPL, line 70: I've already started working towards cleaning up this part in #134, however now that I think about it, it would probably be better for maintenance if the settings file doesn't get any additional modifications in the process of building containers (ie. less mental load on maintainers who have to remember to ensure any changes to the settings in the `Dockerfile` aren't breaking anything in the core settings).

etiaro: I think the environment variables approach is the most flexible, as it will fit well with non-docker installations.
I already updated worflow in #160

etiaro: Thanks, based on deletions in #134 I assume that this fragment of Dockerfile mostly redundant. It's gonna save some time

Dockerfile
```diff
@@ -63,34 +63,25 @@ RUN pip3 install -r requirements_static.txt --user
...
-        s/RUN_LOCAL_WORKERS = True/RUN_LOCAL_WORKERS = False/g;\
-        s/AVAILABLE_COMPILERS = SYSTEM_COMPILERS/#AVAILABLE_COMPILERS = SYSTEM_COMPILERS/g;\
-        s/DEFAULT_COMPILERS = SYSTEM_DEFAULT_COMPILERS/#DEFAULT_COMPILERS = SYSTEM_DEFAULT_COMPILERS/g;\
-        s/USE_UNSAFE_EXEC = True/USE_UNSAFE_EXEC = True/g;\
-        s/#DEFAULT_SAFE_EXECUTION_MODE/#DEFAULT_SAFE_EXECUTION_MODE/g;\
-        s/#USE_UNSAFE_CHECKER = True/#USE_UNSAFE_CHECKER = False/g;\
-        \$afrom basic_settings import *\nALLOWED_HOSTS = ALLOWED_HOSTS + \\['oioioi', '127.0.0.1', 'localhost', 'web'\\]" \
-        settings.py && \
-    cp /sio2/oioioi/oioioi/selenium_settings.py selenium_settings.py && \
+RUN cp /sio2/oioioi/oioioi/selenium_settings.py selenium_settings.py && \
```
twalen, line 84: We can remove 'selenium_settings.py' & './docker-compose-selenium.yml'

oioioi/default_settings.py
```diff
@@ -31,15 +35,30 @@
 # 'uwsgi' - uwsgi daemon
 # 'uwsgi-http' - uwsgi deamon with built-in http server
 # None - nothing will be run
-SERVER = None
+SERVER = os.getenv('OIOIOI_SERVER_MODE', None)
 
+DATABASES = {
+    'default': {
+        'ENGINE': os.getenv('OIOIOI_DB_ENGINE', 'django.db.backends.sqlite3'),
+        'NAME': os.getenv('OIOIOI_DB_NAME'),           # Or path to database file if using sqlite3.
```
twalen, line 43: can we add default values here?

oioioi/deployment/basic_settings.py.template
twalen, line 1: do we need this file?

oioioi/deployment/create_config.py
twalen, line 1: let's do not touch this during this PR, but maybe we can get rid of this script as well? it is used in DockerFile, but maybe the deployment can be simplified

twalen (changes requested):
Looks good, I've added small comments, but once they are resolved we can merge this..

Dockerfile
LiquidPL: This is probably not the PR to do it, but I suppose we could also then remove the entire Selenium test suite in `oioioi_selenium` as well.

oioioi/deployment/basic_settings.py.template
LiquidPL: This is currently used by the Docker setup to let administrators override default configuration values. Since this PR aims to allow configuring the app via environment variables, this should be safe to remove.

oioioi/deployment/create_config.py
LiquidPL: It is also being used for the [non-Docker deployment](https://github.com/sio2project/oioioi/blob/master/INSTALL.rst). I'm not sure how production sio2.mimuw or szkopuł are deployed, but on my own installation i have a `deployment` directory created by this script:

```console
oioioi@5798456b95b5:/sio2/deployment$ ls -l
total 912
-rw-r--r--  1 oioioi oioioi    465 Aug 29  2019 apache-site.conf
drwx------  4 oioioi oioioi   4096 Jan 23 06:58 cache
drwxr-xr-x  1 oioioi oioioi   4096 Apr  2 10:06 logs
-rwxr-xr-x  1 oioioi oioioi    538 Aug 29  2019 manage.py
drwxr-xr-x  6 oioioi oioioi   4096 Aug 29  2019 media
-rw-r--r--  1 oioioi oioioi    495 Aug 29  2019 nginx-site.conf
drwxr-xr-x  1 oioioi oioioi   4096 Jan 23 06:58 pidfiles
-rw-r--r--  1 oioioi oioioi  19361 Nov 14  2020 settings.py
-rw-r--r--  1 oioioi oioioi   7004 Jan 23 06:57 settings.pyc
-rw-r--r--  1 oioioi oioioi 847872 Apr  3 20:34 sioworkersd.db
-rwxr-xr-x  1 oioioi oioioi    109 Aug 29  2019 start_supervisor.sh
drwxr-xr-x 36 oioioi oioioi   4096 Mar  8 12:04 static
-rw-r--r--  1 oioioi oioioi   6382 Aug 29  2019 supervisord.conf
-rw-r--r--  1 oioioi oioioi   1491 Aug 29  2019 wsgi.py
```
Soime of these files are not really going to be modified by end users, others are unnecessary (like apache/nginx configs, for which an example configuration could be added to the documentation). I think we should do away with the whole `oioioi-create-config deployment` folder, and consider something like storing the settings file in a secure location (something like `/etc/oioioi/settings.py`) restricted to a single purpose oioioi user, since the file contains secrets that should be stored securely.

twalen: OK, my mistake, this looks like bigger issue, please ignore my comment.

Dockerfile
etiaro: Will be done in #185.
For now, I just removed selenium from Dockerfile

oioioi/deployment/basic_settings.py.template
etiaro: Looking at the docker-compose.yml, the basic_settings is mounted from repository anyway
```
volumes:
      - $OIOIOI_CONFIGDIR/basic_settings.py:/sio2/deployment/basic_settings.py
```
So generating an empty file here seems useless

etiaro: Okay, removed the basic_settings also from docker-compose.

twalen (commented):
please rosolve the merge conflict and we can merge it

twalen (approved):

PR #180, opened 2023-03-29
Add stop command to easy_toolbox

etiaro:
docker-compose stop is useful for restarting non-debug server to reload settings change, for notifications configuration (which requires to reload settings.py inside of worker container)


twalen (approved):
OK

PR #184, opened 2023-04-03
Teachers admin part 2

Aszanat:


oioioi/teachers/views.py
```diff
@@ -50,6 +52,18 @@ def is_teacher(request):
 def is_not_teacher(request):
     return not_anonymous(request) and not request.user.has_perm('teachers.teacher')
 
+#nowa rzecz, do autocomplete wyszukiwania, czy to przejdzie?
+def get_usernames(request):
```
twalen, line 56: you can use similar approach as in friend_hints_view in particular you can use oioioi.base.utils.user_selection.get_user_hints_view

oioioi/urls.py
```diff
@@ -44,7 +44,8 @@
             urls_module = import_module(app + '.urls')
             if hasattr(urls_module, 'urlpatterns'):
                 urlpatterns += getattr(urls_module, 'urlpatterns')
-        except ImportError:
+        except ImportError as e:
+            print(app, e)
```
twalen, line 48: Please remove this change from this PR, we should resolve this issue in separate Pull Request

oioioi/teachers/views.py
```diff
@@ -50,6 +52,10 @@ def is_teacher(request):
...
```
twalen, line 56: this view should be protected and available only for admin:
```@enforce_condition(is_superuser)```

twalen (approved):
Tested locally, everything works as expected.

Issue #155, opened 2023-03-09
Merge oioioi and szkopuł repositories


edokimok:
# Tasks overview
- [ ] move and adjust commits specific to szkopuł to oioioi repository
  - [x] pick ready-to-merge commit IDs
  - [x] pick commit IDs to delete
  - [ ] adjust rest of commits to fit oioioi repo

### *List of commits on szkopuł only*

- [x] * f64b36aa (HEAD -> master, origin/master, origin/HEAD) dodanie brakującego pliku
- [x] * 08d1a92b add "Projekt MAP" banner
- [x] * 78228e3c Add link to Szkopuł kursy page.
- [x] * 83d76b24 HOTFIX for duplicates in contest ranking (by Alicja Kluczek)
- [ ] * f4d0312a PRODFIX set c++17 for non-makefile sinol packages
- [x] * d268e04d (temporary) Allow lorys to manage packages inside his contest
- [ ] * d46ffb64 (to-master) usercontests should not be public
- [x] * f08f0a88 [tmp] instrument The Queries for perf measure
- [ ] * 3f55b321 [PRODFIX] Change oioioi/notifications/server/npm-shrinkwrap.json
- [x] * d1302cc7 Added Main importer
- [x] * 6fd48ce5 [TEMPFIX] Print where the empty controller hack is needed
- [x] * 8a099a68 TEMPFIX
- [ ] * d7585118 TMPHACK
- [x] * d65532b0 [DEBUG] Added more workers debug
- [x] * 07d0d491 [DEBUG][IMPORTANT] Add some debug messages to receive_from_workers
- [ ] * 0954e73e Uncommited changes
- [x] * 37ba2091 [HOTFIX] Some localchanges to sharingcli on Szkopuł
- [x] * 36d0433d [HOTFIX] Remove oi dependency from participants migration

etiaro:
# determine how to make new settings safe

Move as much of settings.py into basic_settings.py
Move all szkopuł-specific settings into szkopul_settings.py (without configs like filetracker,sentry,db, etc.)

In effect make settings.py as short as possible

sommbre:
Commits ready to merge:
* 83d76b2 HOTFIX for duplicates in contest ranking (by Alicja Kluczek)

Commits to be entirely deleted from szkopuł:
* 6fd48ce [TEMPFIX] Print where the empty controller hack is needed
* d65532b [DEBUG] Added more workers debug
[...]

PR #195, opened 2023-04-20
Update notifications server settings and npm packages

sommbre:
Part of szkopul cleanup (#155).

`npm-shrinkwrap.json` was simply updated with `npm update`.
Changed default settings to make notifications server work with RabbitMQ in docker.

I tested it manually and using tests included in the module.

### DIFF
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -22,6 +22,7 @@ services:
         - "web:127.0.0.1"
     ports:
       - "8000:8000"
+#      - "7887:7887"
     volumes:
       - .:/sio2/oioioi
 #      - ./deployment:/sio2/deployment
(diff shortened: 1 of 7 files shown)


docker-compose-dev.yml
```diff
@@ -22,6 +22,7 @@ services:
         - "web:127.0.0.1"
     ports:
       - "8000:8000"
+#      - "7887:7887"
```
twalen, line 25: is there any reason to have this comment?

A-dead-pixel: That port needs to be exposed for the notifications to work under docker.
I think this can be left here if the purpose is clarified.

sommbre: Added comments to clarify. I decided to leave port 7887 commented as notifications server still requires additional, manual setup (downloading node.js etc.)

twalen (approved):

PR #200, opened 2023-04-20
Szkopul features backporting

sommbre:


twalen (approved):

sommbre:
After szkopul deployment add the following to settings.py:

```
# This is a legacy option for szkopul backwards compatibility.
# Shouldn't be changed unless you know what you are doing.
# Languages added here will use DEFAULT_COMPILERS for non-makefile sinol packages.
OVERRIDE_COMPILER_LANGS = ['cpp']

# Memory limit for input generator job.
# This is a legacy option for szkopul backwards compatibility.
# Shouldn't be changed unless you know what you are doing.
INGEN_MEMORY_LIMIT = 512 * 1024

# This is a legacy option for szkopul backwards compatibility.
# Shouldn't be changed unless you know what you are doing.
# If set to True, usercontests will behave like teachercontests
# and will not be listed on the main page.
# Changing this setting is reversible.
HIDE_USERCONTESTS = True
```

twalen:
I've added above settings to szkopuł settings.

PR #202, opened 2023-05-03
Remove OI dependency from participants migration

sommbre:
Part of szkopul backporting (#155)


A-dead-pixel:
Did you test this on a database that already had these migrations and the oi module on?

twalen (approved):

PR #203, opened 2023-05-06
migration to python 3.8

twalen:


A-dead-pixel:
I once got all tests (iirc) to pass on python 3.9 locally. I will be able to help when I get home next week.
FYI moving to 3.8 doesn't require any work.

twalen:
> I once got all tests (iirc) to pass on python 3.9 locally. I will be able to help when I get home next week. FYI moving to 3.8 doesn't require any work.

it seems that changes in filetracker are required (it contains pinned versions of gevent & greenlet that are incompatible with python 3.9)

DietPawel:
Just for oioioi testing purposes 
> it seems that changes in filetracker are required (it contains pinned versions of gevent & greenlet that are incompatible with python 3.9)
One could ignore gevent as it is needed only for the server part as long as the server runs on older python version. I might try and test.

PR #211, opened 2023-05-18
Update of django

etiaro:


etiaro:
For some reason evalmgr tests hangs

PR #213, opened 2023-05-21
Add ruff as a linter with basic rules

maclick:
Created ruff configuration that uses these rules:
 - pycodestyle
 - pyflakes
 - isort

Also created github workflow for running ruff, as this is really light workflow, I've configured it to run on every push


metenn:
I don't know why nobody has commented on this yet, I think Ruff is a great step forward!
Was Ruff used to format the codebase within this PR already? I can see a ton of changes in at least two commits. Maybe it'd be better if the pass was done in a separate PR, or in a single, final commit, so this PR could be rebase-merged to keep future uses of git-blame more manageable?

Issue #206, opened 2023-05-15
Autocomplete search fields to add


Aszanat:
There might be a few search fields on the site that would be, much or less but still, more useful if they were automatically suggesting names based on already inserted letters, so far I've found:

- Contest rights: add contest permission (Username, must be a teacher)

To see how autocomplete search fields were implemented so far, check "Add teacher" (admin's view) and oioioi/teachers, files:
- admin.py (import: AddTeacherForm and its usage)
- forms.py (again, AddTeacherForm and imports: reverse and UserSelectionField)
- urls.py (re-path of user-search)
- views.py (get_non_teacher_names and imports: get_user_hints_view and JsonResponse)

PR #217, opened 2023-05-25
Issue #206 solve: Autocomplete search fields

Aszanat:
I reorganized admin.py and forms.py in contests because I found a form in admin.py
Also, I added autocomplete search field in a place found in the issue, works as intended.


A-dead-pixel:
I would also add autocompletion to the form for adding new participants to a contest.
Additionally, I don't think we should filter out non-teachers, as they are often non-existent (like on sio2.mimuw, basing on the lack of "Request teacher account" in the user menu) and even if enabled, probably not all contest rights (especially observer ones) would be given to teachers only.

PS.
I think you accidentally deleted ``.gitignore``.

.gitignore
twalen, line 1: please remove this file from PR

oioioi/contests/views.py
```diff
@@ -695,3 +695,7 @@ def reattach_problem_confirm_view(request, problem_instance_id, contest_id):
         'contests/reattach_problem_confirm.html',
         {'problem_instance': problem_instance, 'destination_contest': contest},
     )
+
+def get_teacher_names(request):
```
twalen, line 699: this view needs to be protected (i.e. only for admins)

twalen (changes requested):
remove .gitignore
fix permission issues

twalen:

PR #226, opened 2023-06-08
Django 4.2 migration

etiaro:
Update of Django and other packages


setup.py
```diff
@@ -20,56 +20,187 @@
...
+"django-appconf==1.0.5"
+"django-compressor==4.3.1"
+"django-debug-toolbar==4.1.0"
+"django-extensions==3.2.3"
+"django-formtools==2.4.1"
+"django-gravatar2==1.4.4"
+"django-js-asset==2.0.0"
+"django-libsass==0.9"
+"django-mptt==0.14.0"
+"django-nested-admin==4.0.2"
```
A-dead-pixel, line 57: I don't think this is right (pasted by mistake?):
- no commas
- there are two ``requirements``
- indirect dependencies are included

setup.py
```diff
@@ -20,56 +20,187 @@
...
+    "Django==4.2.2",
+    "pytz==2023.3",
+    "SQLAlchemy==2.0.15",
+    "beautifulsoup4==4.12.2",
+    "PyYAML==6.0",
+    "python-dateutil==2.8.2",
+    "django-two-factor-auth==1.15.2",
+    "django-formtools==2.4.1",
+    "django-registration-redux==2.12",
+    "Celery==4.4.7",    # 5.0 is breaking
```
A-dead-pixel, line 165: Can we please freeze just to major versions?
Exact constraints are really troublesome for some (exotic) deployments.

A-dead-pixel (changes requested):

setup.py
etiaro: Yep, confirmed the fragment was mistaken

A-dead-pixel (changes requested):

etiaro:
@A-dead-pixel okay, the versions are now relaxed as you suggested

A-dead-pixel (approved):

twalen (approved):

Issue #181, opened 2023-03-29
Remove deprecated deployment methods


etiaro:
Remove no longer supported ways/scripts/tutorials of development/production deployment.

Example of such deprecated deployment method can can be found in extra/easy-install folder

Issue #157, opened 2023-03-09
Refactor the documentation


Aleshkev:
These are all the potential sources of SIO2 documentation relating to OIOIOI I could find as of now.

- https://oioioi.readthedocs.io/en/latest/ \
  how to pull request and to review, OIOIOI backend reference (core apps only), architecture overview
- https://github.com/sio2project/oioioi/blob/master/rst/README \
  how to modify oioioi.readthedocs.io
- https://sio2project.mimuw.edu.pl/display/DOC/Preparing+Task+Packages \
  short guide on creating problem packages
- https://github.com/sio2project/oioioi/wiki \
  long guide on creating problem packages (with Polish version)
- https://github.com/sio2project/task-package-examples \
  folder with example problem packages (referenced by oioioi/wiki)
- https://github.com/sio2project/oioioi/tree/master/oioioi/sinolpack/files \
  folder with example problem packages
- https://github.com/sio2project/oioioi#readme \
  ~~how to install OIOIOI~~ (outdated), Docker scripts (outdated?), how to run Selenium tests
- https://github.com/sio2project/oioioi/blob/master/CONTRIBUTING.rst \
  link to Github's guide to pull requests and review, how to do Transifex translations
- https://sio2project.mimuw.edu.pl/display/DOC/SIO2+Translator%27s+Guide \
  ~~how to do Transifex translations~~ (legacy site)
- https://github.com/sio2project/oioioi/blob/master/GUIDE.md \
  how to install OIOIOI, how to run tests (Selenium?, Cypress), looks informal
- https://github.com/sio2project/oioioi/blob/master/oioioi_selenium/README.rst \
  how to run Selenium tests, how to create and fix Selenium tests
- https://github.com/sio2project/oioioi/blob/master/INSTALL.rst \
  ~~how to install OIOIOI~~ (very outdated)
- https://github.com/sio2project/oioioi/blob/master/UPGRADING.rst \
  guide to upgrading OIOIOI deployments (with a rather long guide how to upgrade from very old versions?)
- https://sio2project.mimuw.edu.pl/display/DOC/Home \
  ~~random outdated stuff, probably unimportant~~

Our Django apps seem not to be documented beyond a one-sentence description of what they do, except these:
- https://github.com/sio2project/oioioi/blob/master/oioioi/notifications/README.rst \
  how to run, use, and test the server for instant notifications
- https://github.com/sio2project/oioioi/tree/master/oioioi/testspackages#readme \
[...]

PR #229, opened 2023-06-25
Remove selenium, easy_install

A-dead-pixel:
Closes https://github.com/sio2project/oioioi/issues/181.
Resolves https://github.com/sio2project/oioioi/issues/185.


Xenox40:
It looks like `db_init.sh` is only used by selenium build and it uses `oioioi_selenium/data.json`. Maybe there are other files like this one.

I'd consider leaving some information from [INSTALL.rst](https://github.com/sio2project/oioioi/pull/229/files#diff-ce342037df7e6de102c3157b708a389b128142aae879421094c71d9f97f5084c). Things like what packages are used by SIO or how to setup nginx with uwsgi can still be useful.

A-dead-pixel:
You were right, I found that OIOIOI_CONFIGDIR isn't used anymore (easy-install needed it) , so I removed it from the README.
I also left INSTALL.rst alone, as dealing with it properly is probably a job for another PR, maybe related to #157.

twalen (approved):

A-dead-pixel:
Note: #110 will become irrelevant after this is merged.

PR #235, opened 2023-07-10
Fixup the settings refactor

A-dead-pixel:
- since default_settings.py don't provide a working email configuration, sending account activation emails should be disabled by default. Without this, one can't register new accounts on a dev docker setup.
~~- OIOIOI_SERVER_MODE needs to be set to 'uwsgi-http' at least for dev deployments, otherwise no server is started.~~
- the default value for the SERVER setting should be a string instead of None, since there is no way to unset an env var in docker. With this change one can freely override a value from a Dockerfile in docker-compose.
- None is not a valid TIME_ZONE setting (at least in docker)
- if OIOIOI_TIMEZONE is to be of any use, we shouldn't override TIME_ZONE in settings.py.template by default

### DIFF
--- a/GUIDE.md
+++ b/GUIDE.md
@@ -30,7 +30,7 @@ In order to use `easy_toolbox.py` alternative method, check python package requi
 
 To build OIOIOI image run 
 ```bash
-OIOIOI_UID=$(id -u) docker-compose -f docker-compose-dev.yml -f extra/docker/docker-compose-dev-noserver.yml build
+OIOIOI_UID=$(id -u) docker-compose -f docker-compose-dev.yml build
 ```
 or
 ```bash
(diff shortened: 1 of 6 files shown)


Xenox40:
>    OIOIOI_SERVER_MODE needs to be set to 'uwsgi-http' at least for dev deployments, otherwise no server is started.

I think this is intended behavior (None). It's just easier to manully run a server manually, see the logs in real time and you can re-run it were fast as you want.

Two more things to fix in  `README`: 
- Dev installation still has the old `extra/docker/docker-compose-dev-noserver.yml` in docker-compose commands.
- The line `id=$(docker create oioioi-dev)` is broken, the name should be: `sio2project/oioioi-dev`, at least I had to change it.

A-dead-pixel:
I addressed all of your suggestions, thanks for the feedback!

> * The line `id=$(docker create oioioi-dev)` is broken, the name should be: `sio2project/oioioi-dev`, at least I had to change it.

Seems like a leftover from 628efee5.

twalen (approved):

twalen (approved):

PR #243, opened 2023-08-09
Upstream some of our production only changes 2018-2023

DietPawel:
Currently I am cleaning up production code and found several commits which are not on the master branch.

I would prefer to merge them all at once once I finish processing them.

In case of any doubts please comment under particular commits.


DietPawel:
Please keep in mind that checks, which failed are the same that failed on the [master branch](https://github.com/sio2project/oioioi/actions/runs/5887103370/job/15965961618).

oioioi/notifications/processors.py
```diff
@@ -20,6 +21,7 @@ def generate_token():
     return new_token
 
 
+@request_cached
 def get_notifications_session(session):
```
Wolf480pl, line 25: the argument of this function is a session, not a request - are you sure `@request_cached` decorator will work correctly on it?

oioioi/liveranking/templates/liveranking/autoDonuts.html
```diff
@@ -69,8 +92,11 @@
 </head>
 <body>
     <div id="header"></div>
-    <div id="donut1"></div>
-    <div id="donut2"></div>
+    <div id="info">
+        <div id="text"></div>
```
Wolf480pl, line 96: is this for the trivia? :thinking:

DietPawel: Exactly, it was mentioned in the original commit "add a #text div in autoDonuts, which will be useful for displaying trivia"

oioioi/sinolpack/utils.py
```diff
@@ -1,7 +1,42 @@
...
+                 .parsed_config
+    except ExtraConfig.DoesNotExist:
+        config = {}
+
+    lang = environ.get('language')
+    for _, job in environ.get('tests', {}).items():
+        extra_args = config.get('extra_execution_args', {}).get(lang, [])
+        if isinstance(extra_args, str):
+            extra_args = [extra_args]
+        if extra_args:
```
Wolf480pl, line 22: if extra_args is `""` initially, then it will be transformed to `[""]`  which is truthy so it will be propagated.
It's an edge case, and I don't even know if it will break anything downstream, and it looks like
```
extra_execution_args:
  c:
```
results in None, not "" - you'd have to explicitly write
```
extra_execution_args:
  c: ""
```
which someone could conceivably write with the intention of adding an empty argument, so I guess we can leave it as is, though in an ideal world I'd want to see a comment in the code explaining why this is ok.

oioioi/liveranking/templates/liveranking/autoDonuts.html
Wolf480pl: is the trivia code still in, or was it axed? I liked that feature but I think we lost the list of trivia messages :/

oioioi/notifications/processors.py
DietPawel: Indeed it is in violation with request_cached docstring. To be honest I saw a commit adding a bunch of `request_cached` docorators and knowing it was operational on production made no further digging.

I think it would be best to remove this change here, and bring later it back provided it is safe and beneficial.

oioioi/sinolpack/utils.py
DietPawel: I see the point. It could use some touch ups and reformatting for sure.

oioioi/quizzes/controllers.py
```diff
@@ -222,6 +222,8 @@ def render_report_failure(self, request, report):
         return ProblemController.render_report(self, request, report)
 
     def render_report(self, request, report):
+        if report.kind == u'FAILURE':
+            return super(QuizProblemController, self).render_report(request, report)
         problem_instance = report.submission.problem_instance
         if report.kind == 'FAILURE':
```
Wolf480pl, line 233: we have the same if here wit ha different action - are you sure this commit wasn't upstreamed previously, and then mis-rebased?

oioioi/notifications/processors.py
Wolf480pl: yeah, the rest of the commit looks fine but this one line looks like it may've been mis-rebased due to a refactor, better skip it for now

oioioi/quizzes/controllers.py
```diff
@@ -198,15 +198,15 @@ def _submit_answers(self, selected_answers, question, submission):
...
     def _get_selected_answers(self, form_data, field_id, question):
-        field_value = form_data.get(field_id)
+        field_value = form_data.get(field_id, '')
         if question.is_text_input:
             if isinstance(field_value, list):
                 return field_value[:1]
             return [field_value]
         elif question.is_multiple_choice:
-            return [int(a) for a in field_value]
+            return [int(a) for a in filter(bool,field_value)]
```
Wolf480pl, line 207: that will reject zero - are you sure zero is never an answer ID?

Wolf480pl: oh, nvm you do that before int(a), it's ok

oioioi/liveranking/templates/liveranking/autoDonuts.html
DietPawel: Since trivia code got lost so will be this change

oioioi/quizzes/controllers.py
DietPawel: Indeed, I will remove that commit.

oioioi/notifications/processors.py
```diff
@@ -1,6 +1,7 @@
 import random
 import string
 
+from oioioi.base.utils import request_cached
```
Wolf480pl, line 4: unused import?

Wolf480pl (approved):

Wolf480pl (dismissed):
Looks good to me, but I don't know current sio2 development practices, so I don't know if and when this should be merged.

oioioi/notifications/processors.py
DietPawel: fixed!

Wolf480pl (dismissed):

oioioi/oireports/views.py
Wolf480pl, line 1: I think it'd be worth double-checking if:
- generating PDF/XML reports works without this
- this doesn't cause an FD leak

oioioi/oireports/views.py
```diff
@@ -167,7 +167,6 @@ def _serialize_report(user, problem_instances, test_groups):
                 codefile=source_file.file.name,
             )
         )
-        source_file.close()
```
Wolf480pl, line 170: I think it'd be worth double-checking if:
- generating PDF/XML reports works without this
- this doesn't cause an FD leak

DietPawel:
@twalen is it fine to rebase now?

oioioi/oireports/views.py
DietPawel: will check later

DietPawel: when reverting this change generating PDF report throws
```
AttributeError: '_FileLikeFromResponse' object has no attribute 'close'
```

DietPawel: It would make more sense to have `source_file.file.close()`. After digging on the issue, django model's File can open itself implicitly when reading. It should be closed only it it was explicitly opened. I decided not to put more time into it now, as it seems to be fine and unfortunately other commits require much more time and work.

oioioi/sinolpack/package.py
```diff
@@ -1139,9 +1139,11 @@ def _prepare_overrides(self, lang):
...
-                time_limit=test.time_limit,
-                memory_limit=test.memory_limit,
+            LanguageOverrideForTest.objects.update_or_create(
+                defaults={
+                    'time_limit': test.time_limit,
+                    'memory_limit': test.memory_limit,
+                },
                 test=test,
                 language=lang,
             )
```
A-dead-pixel, line 1149: This duplicates #228, which also covers the bug in a test.

DietPawel: I see, it's a pity two people had to come up with solution to the same problem.
I would be happy to include your tests with this change. Are You fine with that?

DietPawel:
As master has changed I rebased this branch and decided to include @A-dead-pixel 's  tests
Here are the changes: https://github.com/sio2project/oioioi/compare/ef63dff..144835b1

oioioi/sinolpack/package.py
A-dead-pixel: Yeah, it's fine.

twalen (approved):

DietPawel:
@twalen As i asked here https://github.com/sio2project/oioioi/pull/243#issuecomment-1682420601 can we make a non-squashed merge?

PR #245, opened 2023-08-11
Fix class name shadow braking user change form in admin panel

### LINKED ISSUE #244: Reset password link is missing in admin's user change form
The default [`UserChangeForm`](https://docs.djangoproject.com/en/1.8/topics/auth/default/#django.contrib.auth.forms.UserChangeForm) in django has a link to reset password form. It's useful when students forget thier passwords and internet connection is not available. The current form has a simple text input with hash string instead.
The endpoint `/admin/auth/user/{id}/password` works fine.

How it looks now:
image

How it should look:
image

Xenox40:
Closes #244.

This small change fixes the issue by renaming the class responsible for enforcing password confirmation when user wants to change its email address.

The `UserChangeForm` class introduced in https://github.com/sio2project/oioioi/commit/363ce7601719240a532cbe0585b6d57eaa70d393 https://github.com/sio2project/oioioi/blob/40a377d3f2d5cd9c94d01f03e197501ce4aab597/oioioi/base/forms.py#L235-L257
shadows the django one imported earlier https://github.com/sio2project/oioioi/blob/40a377d3f2d5cd9c94d01f03e197501ce4aab597/oioioi/base/forms.py#L8-L12 
which is used as a base for admin's user change form https://github.com/sio2project/oioioi/blob/40a377d3f2d5cd9c94d01f03e197501ce4aab597/oioioi/base/forms.py#L266-L270

### DIFF
--- a/oioioi/base/forms.py
+++ b/oioioi/base/forms.py
@@ -232,7 +232,7 @@ class UserForm(forms.ModelForm):
         PreferencesSaved.send(self, user=instance)
         return instance
 
-class UserChangeForm(UserForm):
+class OioioiUserForm(UserForm):
     confirm_password = forms.CharField(widget=forms.PasswordInput(), required=False)
 
     class Media(object):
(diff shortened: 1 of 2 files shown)


DietPawel:
I rerun the tests, which pass in my local testing, just not here. This is simple change, which fixes a bug. I would appreciate merging it!

twalen (approved):

DietPawel:
Thank You for quick reaction!

PR #248, opened 2023-08-21
Implement easy import of schools from the ME

Igifigi:
This PR implements an algorithm that will bring order to school selection in SIO. Implemented script is used to import schools from the RSPO database, which is provided by the Ministry of Education and contains all educational institutions in Poland. Due to the poor quality of this data, it is likely that SIO administrators will modify school data. The task of this script is to detect which corrections from newer and more recent versions of the RSPO database should be accepted and applied, and which should not.


twalen:
some more verbose commit messages might be useful (instead of "temp commit")

Igifigi:
> some more verbose commit messages might be useful (instead of "temp commit")

As you can see, this PR is set as a draft, so commit names will be changed soon (probably tomorrow).

Igifigi:
@twalen not it's ready for review :))

DietPawel:
I am missing some migration related trickery to allow migrating a database with existing schools lacking rspo fields.
As well as I am not sure it is right to get rid of legacy non-rspo csv import.

oioioi/oi/management/commands/import_schools.py
```diff
@@ -1,122 +1,533 @@
...
-                    setattr(school, column, row[column])
+                # If the MoE has made changes that we have not, then accept MoE changes.
+                if not db_and_curr_the_same and db_and_last_the_same:
+                    self.log(
+                        f'Note: The school [rspo: {db_school.rspo}] has been changed in the RSPO database. The changes have been applied.',
+                        school_before=db_school,
+                        school_after=translate_dict_school(curr_rspo_school),
+                    )
+                    update_school(curr_rspo_school, db_school)
+                    rows_affected += 1
```
MasloMaslane, line 451: Changes from MoE may introduce new duplicates.
```suggestion
                    try:
                        with transaction.atomic():
                            update_school(curr_rspo_school, db_school)
                        rows_affected += 1
                    except utils.IntegrityError:
                        self.log(
                            f'Warning: Django found a duplicate school. Please manually check the school {db_school.rspo} data in SIO database, '
                            'current RSPO and the last backup. Something suspicious may be going on.'
                        )
```

Iteron-dev:
The changes from this PR have been mostly incorporated here: 9e084448da629b51bdcdcb95c1521ceff3028121 – from what I’ve seen, only `oioioi/oi/management/commands/import_schools_legacy.py` isn’t working correctly. Therefore, the non-rspo CSV import is not supported – the question is whether it’s worth fixing?

Iteron-dev:
> From our perspective legacy import is not needed. And since RSPO database is public, I think the command can be deleted

Sure - I'll create a PR that removes the legacy import.

PR #249, opened 2023-08-21
Change school_select_form selects to select2

Igifigi:
A minor change has been made in this PR to make it easier to select a school in the contest registration form. All <select> elements have been changed to <select2> elements, which will allow you to select your educational institution more efficiently,


DietPawel:
I think we should be aware of dependencies in critical system components like registration. In case of offline use CDN might not be available and select2 should be provided via alternative source.

Igifigi:
will be fixed soon, for now i set as draft

oioioi/oi/static/oi/select2.min.js
A-dead-pixel, line 1: Imho it would be nice if these files were in `base` instead of `oi`, as they will probably get used in other modules too and the `oi` module is often disabled.

Igifigi: ok

Iteron-dev:
Resolved in https://github.com/sio2project/oioioi/commit/9e084448da629b51bdcdcb95c1521ceff3028121.

PR #252, opened 2023-08-30
Change 'registration' to 'sign-up' in the context of registration in SIO

Igifigi:
SIO users often did not distinguish between contest registration and system registration. For this reason, registration for the SIO was changed to sign-up.

### DIFF
--- a/oioioi/base/registration_backend.py
+++ b/oioioi/base/registration_backend.py
@@ -7,7 +7,7 @@ from django.contrib.auth.models import User
 from django.contrib.auth.views import PasswordResetView, PasswordResetDoneView
 from django.contrib.sites.requests import RequestSite
 from django.urls import re_path, reverse_lazy
-from django.views.generic import TemplateView
+from django.views.generic import TemplateView, RedirectView
 from registration import signals
 from registration.backends.default.views import (
     RegistrationView as DefaultRegistrationView,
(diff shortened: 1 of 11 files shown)


oioioi/base/registration_backend.py
```diff
@@ -63,18 +63,18 @@ def register(self, form):
...
 urlpatterns = [
-    re_path(r'^register/$', RegistrationView.as_view(), name='registration_register'),
+    re_path(r'^sign-up/$', RegistrationView.as_view(), name='sign-up'),
 ]
 
 if not settings.SEND_USER_ACTIVATION_EMAIL:
     urlpatterns += [
         re_path(
-            r'^register/complete/$',
+            r'^sign-up/complete/$',
```
DietPawel, line 72: I would like to see some redirection from previous register to sign-up just not to broke already existing links in some content

oioioi/base/registration_backend.py
```diff
@@ -63,18 +63,18 @@ def register(self, form):
 
 
 urlpatterns = [
-    re_path(r'^register/$', RegistrationView.as_view(), name='registration_register'),
+    re_path(r'^sign-up/$', RegistrationView.as_view(), name='sign-up'),
```
DietPawel, line 66: I am not sure it is bad, just thinking if `sign-up` is unique enough

DietPawel (changes requested):

oioioi/base/registration_backend.py
A-dead-pixel: I think that `create-account` will be better and sufficient. Also, then the Polish translation would be different from "rejestracja".

DietPawel: I would like to say I only disliked url's name, not the url itself. As for the name I like *create account*, as it contains a noun making it more specific. In terms of translations which is the main point of this PR I consider "załóż konto" as good translation of "sign-up". Anything that won't be the same two words for making an account and enrolling to the contest will fix it for me.

oioioi/base/registration_backend.py
```diff
@@ -63,17 +63,23 @@ def register(self, form):
...
+    re_path(r'^register/$', RedirectView.as_view(url='/sign-up/', permanent=True), name='registration_register'),
```
DietPawel, line 67: Again, I am picky, but `registration_redirect` is so much better than `registration_register` in my opinion

oioioi/base/registration_backend.py
```diff
@@ -63,17 +63,23 @@ def register(self, form):
...
         re_path(
-            r'^register/complete/$',
+            r'^sign-up/complete/$',
             TemplateView.as_view(
                 template_name='registration/'
                 'registration_and_activation_complete.html'
             ),
+            name='sign-up_complete',
+        ),
+        re_path(
```
DietPawel, line 80: I asked for it, but I must have marked the wrong line, sorry. I like the redirection from `/register/` to `/sign-up/`, but this complete link is a not necessary.

DietPawel (changes requested):

DietPawel (approved):

PR #254, opened 2023-09-02
Allow different ACM penalty time in different contests

DietPawel:
Change-Id: I8aceb796409ad4319012f210868156e24de0b0d1


DietPawel:
One more of our production changes. I missed it previously since it was joint with other changes. #243

metenn (approved):

PR #257, opened 2023-09-03
Add the contest owner permission

### LINKED ISSUE #174: Add a permission type per contest that can add other contest admins.
I'll suggest calling it "contest owner". The contest owner is added by a system admin to the contest, but can add other contest admins to this contest without having to ask the system owner to do it.

A-dead-pixel:
Close #174.

Also make the user field in certain object editing forms default to the username (which needs to be posted) instead of an id.

### DIFF
--- a/oioioi/base/templatetags/check_perm.py
+++ b/oioioi/base/templatetags/check_perm.py
@@ -19,12 +19,7 @@ class CheckPermNode(Node):
             context[self.var] = False
         else:
             user = context['user']
-            if perm == 'contests.contest_basicadmin':
-                context[self.var] = user.has_perm(perm, obj) or user.has_perm(
-                    'contests.contest_admin', obj
-                )
-            else:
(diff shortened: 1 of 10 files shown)


oioioi/base/permissions.py
```diff
@@ -177,3 +177,7 @@ def not_anonymous(request):
 @make_request_condition
 def is_superuser(request):
     return request.user.is_superuser
+
+@make_request_condition
+def isnt_superuser(request):
+    return not request.user.is_superuser
```
metenn, line 183: `Condition` has `__invert__` implemented, so one can probably just use `~is_superuser`?

A-dead-pixel: Yeah, you are right.
I will improve this PR after OI.

A-dead-pixel:
I rebased against current master, addressed the suggestion and improved the tests.

MasloMaslane (approved):

PR #258, opened 2023-09-06
Pass input test name to inwer

MasloMaslane:
This PR introduces passing test name to inwer as an argument. This will allow task creators to write inwers that will be able to verify if tests are in correct groups. It is a backwards-compatible change, because inwers shouldn't expect any arguments.

Related: https://github.com/sio2project/sioworkers/pull/22


twalen (commented):
Looks good, but the changes for sioworkers need be merged first (build is failing)

twalen (approved):
OK

twalen:
But before merging please:
* resolve merge conflict
* make sure that all tests are passing

MasloMaslane:
I had to change [test_interactive_package.tgz](https://github.com/sio2project/oioioi/pull/258/files#diff-3bd1e6c5b8510d9eae9830da19c4e5a6e1e1467fe661822aa4299574fd603523), because if a file was passed as argument to inwer, it tried to open this file. I removed this logic.

geoff128 (approved):

PR #262, opened 2023-09-18
Added password validation

BrutBurger:
According to FRI privacy policy passwords needs to have:
      >min 8 characters
      >contain a lowercase and uppercase letter, as well as digit and special character

This PR covers that issue. 
For it to work, the following needs to be added to the settings.py:

AUTH_PASSWORD_VALIDATORS = [
    {
        "NAME": "django.contrib.auth.password_validation.MinimumLengthValidator",
        "OPTIONS": {
            "min_length": 8,
        },
    },
    {'NAME': 'oioioi.base.validators.NumberValidator', },
    {'NAME': 'oioioi.base.validators.UppercaseValidator', },
    {'NAME': 'oioioi.base.validators.LowercaseValidator', },
    {'NAME': 'oioioi.base.validators.SymbolValidator', },
]

### DIFF
--- a/oioioi/base/validators.py
+++ b/oioioi/base/validators.py
@@ -1,0 +1,75 @@
+import re
+
+from django.core.exceptions import ValidationError
+from django.utils.translation import gettext as _
+from django.contrib.auth import password_validation
+
+class NumberValidator(object):
+    def validate(self, password, user=None):
+        if not re.findall('\d', password):
+            raise ValidationError(
+                _("The password must contain at least 1 digit, 0-9."),
+                code='password_no_number',
+            )
+
+    def get_help_text(self):
+        return _(
+            "Your password must contain at least 1 digit, 0-9."
+        )
+
+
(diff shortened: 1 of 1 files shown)


twalen (approved):
Code looks ok, but the latest password recommendations are focusing on length not type of characters:
https://sekurak.pl/jaka-jest-najczestsza-zla-rada-dawana-w-kontekscie-hasel/
https://support.microsoft.com/en-us/windows/create-and-use-strong-passwords-c5cebb49-8c53-4f5e-2bc4-fe357ca048eb

DietPawel:
@twalen That is exactly what I pointed out. I even gave them an example of `quo8UaQuie3Mie9ahwoo` and `O1impi@da`. It is all about possible combinations and entropy. Enforcing use of special characters lowers no of possible password. Unfortunately making a sensible requirements would mean changing to policy and arguing with GDPR officer.

PR #267, opened 2023-09-20
Fix not accounting for disabled activation emails

A-dead-pixel:
...in the "Sign-up successful" message.

This is another fixup of the Register -> Sign-up switch.
The change in url names resulted in django-registration-redux's view/url not being overriden, so the "Sign-up successful" message would mention activation emails even when sending those was disabled.


MasloMaslane:
On sio2.mimuw we changed it like this:
```diff
diff --git a/oioioi/base/registration_backend.py b/oioioi/base/registration_backend.py
index f38f2286..35d5e4be 100644
--- a/oioioi/base/registration_backend.py
+++ b/oioioi/base/registration_backend.py
@@ -63,7 +63,13 @@ class RegistrationView(DefaultRegistrationView):
 
 
 urlpatterns = [
-    re_path(r'^sign-up/$', RegistrationView.as_view(), name='sign-up'),
+    re_path(
+        r'^sign-up/$',
+        RegistrationView.as_view(
+            success_url=reverse_lazy('sign-up_complete')
+        ),
+        name='sign-up',
+    ),
     re_path(
         r'^register/$',
         RedirectView.as_view(pattern_name='sign-up', permanent=True),
```
I think this solutions is better, as it doesn't require changing `sign-up` -> `register` again.

A-dead-pixel:
What @MasloMaslane suggested seems to be the proper way to do this, though I had to change it a bit for it to work with activation emails enabled.

DietPawel (approved):
Looks good!

twalen (approved):

PR #268, opened 2023-09-21
Allow model solutions with long names

MasloMaslane:
I changed the model solution's regex from `^<task id>[0-9]*([bs]?)[0-9]*` to `^<task id>[0-9]*([bs]?)[0-9]*(sol.*)?`. This will allow new model solutions to have long and descriptive names while not breaking old model solutions (for example `abc1s10sol_very_long_name.cpp`).


twalen (approved):
OK

PR #274, opened 2023-10-06
Address a few warnings

A-dead-pixel:
- `(urls.W005) URL namespace 'contest:oioioiadmin' isn't unique. You may not be able to reverse all URLs in this namespace` - from manage.py
- `RuntimeWarning: DateTimeField Message.date received a naive datetime while time zone support is active.` - from uwsgi/runserver logs
- `SyntaxWarning: "is not" with a literal. Did you mean "!="?` - from at least docker build logs
- captcha check complaining about missing sox and flite - from at least pytest

I don't know of any more visible warnings other than those in logs of celery and filetracker, which should be addressed by updating dependencies outside this PR.
If I missed some, please tell me.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,8 @@ RUN apt-get update && \
         sudo \
         libstdc++6 \
         zlib1g \
+        sox \
+        flite \
         locales \
         python3-pip && \
     apt-get clean
--- a/oioioi/clock/urls.py
+++ b/oioioi/clock/urls.py
@@ -5,5 +5,6 @@ from oioioi.clock import views
 app_name = 'clock'
 
 urlpatterns = [
-    re_path(r'^admin/time/$', views.admin_time, name='admin_time'),
+    # Don't use the 'admin/' prefix, as that sometimes results in breakage.
+    re_path(r'^admin_time/$', views.admin_time, name='admin_time'),
 ]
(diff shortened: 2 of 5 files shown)


oioioi/clock/urls.py
```diff
@@ -5,5 +5,6 @@
 app_name = 'clock'
 
 urlpatterns = [
-    re_path(r'^admin/time/$', views.admin_time, name='admin_time'),
+    # Don't use the 'admin/' prefix, as that sometimes results in breakage.
+    re_path(r'^admin_time/$', views.admin_time, name='admin_time'),
```
MasloMaslane, line 9: This should fix admin time not working while in contest right?

A-dead-pixel: Yes, though fixing the oioioiadmin namespace warning made admin time fail even outside of contests.

Dockerfile
```diff
@@ -22,6 +22,8 @@ RUN apt-get update && \
         sudo \
         libstdc++6 \
         zlib1g \
+        sox \
```
DietPawel, line 25: Nice change!

This is probably the reasons dev's never noticed things like #277 before.

DietPawel (commented):
I have not idea why it is not yet approved.

I did not run the code, but if someone would click test it, I would approve the change.

PR #275, opened 2023-10-09
Forum improvements

MasloMaslane:
Things changed:
- count number of threads and posts in database instead of in django
- Add option to allow only participants to post on forum
- Add pagination in category view


oioioi/forum/utils.py
```diff
@@ -68,6 +69,9 @@ def is_proper_forum(request, *args, **kwargs):
 def can_interact_with_users(request):
     if request.user.is_anonymous:
         return False
+    if request.contest.forum.only_for_registered and \
+            (not is_participant(request) and not is_contest_admin(request)):
+        return False
```
A-dead-pixel, line 74: Imo `is_contest_basicadmin` or `can_admin_contest` would be better here.

MasloMaslane: Many functions in forum app already use `is_contest_admin`, even in this function
```py
    return is_contest_admin(request) or (not is_banned and not is_locked)
```

oioioi/forum/views.py
```diff
@@ -52,9 +53,21 @@
...
-    category_set = request.contest.forum.category_set.prefetch_related(
-        'thread_set', 'thread_set__post_set'
-    ).all()
+    categories = []
+    for category in request.contest.forum.category_set.all():
+        threads = category.thread_set.annotate(
+            post_count=Count('post'),
+            reported_count=Count('post', filter=Q(post__reported=True)),
+        ).all()
+        post_count = sum([thread.post_count for thread in threads])
```
DietPawel, line 62: You are creating a list here just to sum it's all elements. It wastes memory. I suggest to use a [generator expression](https://docs.python.org/3/reference/expressions.html#generator-expressions) like so:
```
sum( (thread.post_count for thread in threads) )
```

oioioi/forum/views.py
```diff
@@ -52,9 +53,21 @@
 @enforce_condition(contest_exists & can_enter_contest)
 @enforce_condition(forum_exists_and_visible & is_proper_forum)
 def forum_view(request):
-    category_set = request.contest.forum.category_set.prefetch_related(
-        'thread_set', 'thread_set__post_set'
-    ).all()
+    categories = []
+    for category in request.contest.forum.category_set.all():
+        threads = category.thread_set.annotate(
```
DietPawel, line 58: Here we are creating a database query for every category. How about querying for number of posts in each category in a single query? I could write an SQL for that, so should be possible to do that in django too.

oioioi/forum/utils.py
DietPawel: I am for consistency. In case of doubts I would suggest to move it to an issue.

PR #280, opened 2023-10-23
Forum improvements and fixes

MasloMaslane:
Changes:
- added pagination in thread view
- fixed counting of posts and threads
- added max post length


DietPawel (approved):
This change I already rolled out in our instance and I approve it.

Just for the record, I think that specifying *from which post id the view should display next x posts* is far superior than using page numbers, but this auto-paginate is so widely used that it is still best approach right here.

twalen (approved):

PR #291, opened 2023-11-19
Change naming of model solutions

MasloMaslane:
Due to a miscommunication with OI's Jury, the naming of model solutions with long names was slightly wrong. I prefer the `_` over `sol` as separator, so I think it's a good change.


oioioi/sinolpack/package.py
```diff
@@ -1248,7 +1248,7 @@ def _get_model_solutions_sources(self):
         """
         lang_exts_list = getattr(settings, 'SUBMITTABLE_EXTENSIONS', {}).values()
         extensions = [ext for lang_exts in lang_exts_list for ext in lang_exts]
-        regex = r'^%s[0-9]*([bs]?)[0-9]*(sol.*)?\.(' + '|'.join(extensions) + ')'
+        regex = r'^%s[0-9]*([bs]?)[0-9]*(_.*)?\.(' + '|'.join(extensions) + ')'
```
twalen, line 1251: Personally I prefer the previous regex, at this moment other utility programs like inwer/ingen/chk are single character apart to match new regex.

MasloMaslane:
Jury would really prefer if it was `_` instead of `sol`. Also, there already exist tests if correct model solutions were collected, for example `TestSinolPackage::test_full_unpack_update`

twalen (approved):
ok

PR #301, opened 2023-12-06
Display % of points outpu by checker

### LINKED ISSUE #288: % of points output by chk to be displayed by SIO
When the checker program (xyzchk) states that a given contestant's output should be given a certain percentage of points, it would be very convenient to write this information automatically in a comment on the report page (see e.g. test 2ocen on https://sio2.mimuw.edu.pl/c/oi31-1/s/630433/). This is especially important in example tests in which this information is otherwise not visible to the contestant, unless a Jury member writes it explicitly in the comment output by the checker program.

SZonkil:
Addition of translation is required

Closes #288


twalen (approved):

oioioi/programs/handlers.py
```diff
@@ -555,6 +555,9 @@ def make_report(env, kind='NORMAL', save_scores=True, **kwargs):
         test_report.score = result['score'] if save_scores else None
         test_report.status = result['status']
         test_report.time_used = result['time_used']
+        percentage = result.get('result_percentage', 0.)
+        if percentage != 100. and percentage != 0.:
```
DietPawel, line 559: Can You please convince me it is okay here to check equality of some floating point number? I would prefer to rather see
```
if percentage > 0 and percentage < 100:
```
keep in mind that sometimes "different" number are equal due to precision
```
>>> 100.000000000000001 == 100.
True
>>> 100.00000000000001 == 100.
False
```

oioioi/programs/models.py
```diff
@@ -326,6 +326,7 @@ class TestReport(models.Model):
     max_score = ScoreField(null=True, blank=True)
     time_used = models.IntegerField(blank=True)
     output_file = FileField(upload_to=make_output_filename, null=True, blank=True)
+    result_percentage = models.FloatField(null=True, blank=True)
```
DietPawel, line 329: Do we want to add some constraint to force the range of result_percentage's values?

oioioi/programs/templates/programs/report-comments.html
```diff
@@ -24,7 +24,22 @@
                     {% endif %}
                 {% endif %}
                 {% if allow_test_comments %}
-                    <span>{{ test.comment }}</span>
+                    <span>
+                    {% if test.comment %}
+                        {% if test.result_percentage %}
+                            {% blocktranslate with test.result_percentage as percentage %}
+                                Granted {{ percentage  }}% points, comment:
```
DietPawel, line 31: How about "Granted {{ percentage }}% *of* points". Also as `percentage` could be 33.33333333 I think we should pretty print this floating point value

oioioi/programs/templates/programs/report-comments.html
```diff
@@ -24,7 +24,22 @@
                     {% endif %}
                 {% endif %}
                 {% if allow_test_comments %}
-                    <span>{{ test.comment }}</span>
+                    <span>
+                    {% if test.comment %}
```
DietPawel, line 28: This logic is quite complex. Do we have any performance degradation due to this change?
I also disagree that the percentage should be graphically the same as comment. They should be visually different.

DietPawel (changes requested):

oioioi/programs/handlers.py
top-sekret: Also, the domain of these values is not necessarily `[0, 100]`. I actually have abused this mechanism with values well above 100 to circumvent some architectural shortcomings once.

DietPawel: Thanks for the insight. In the wiki it says that the percentage is a whole integer. It says that this option can be used to give partial scoring. It does not mention whether it is allowed to grant more than 100%.  It should be more documented in the specs

As for this logic, I am not sure we should tell user that he got 150% for this test, as this i probably not the intended behavior for this feature and was properly called "an abuse".

I suspect there would be disagreements about this, so please express your opinions here!

MasloMaslane: I don't know if the percentage value should be greater than 100, but in the context of this PR I think that the message should not be displayed when the value is 0 (then the result should be WA) or 100 (then there is no comment and you can see assume that you have full points).

But if we decided to allow a percentage greater than 100, we would have to find another way to check whether this value is equal to 100

DietPawel: First of all, if we could change this to an Integer, the float-related discussion is unnecessary. Is there any reason to use `FloatField` not `IntegerField`?

Secondly I think that Verdict and partial scoring are different things and it should be allowed to give 0% points but OK as a verdict. The checker returns both the verdict and partial scoring independently. We should treat is as such independent values.

Thirdly If partial scoring does not affect number of points awarded there should be no comment (what for it would be). If user get less points we definitely should post the comment. If it is allowed to give more, it is debatable if it should be displayed.

DietPawel:
CC: @Kostero @craniac-swistak

oioioi/programs/models.py
twalen: what about using fraction field: https://django-fractions.readthedocs.io/en/latest/readme.html

twalen: It's not perfect (only Decimal fractions).

oioioi/programs/templates/programs/report-comments.html
twalen: @DietPawel do you still request changes?

twalen (dismissed):

oioioi/programs/templates/programs/report-comments.html
DietPawel: There are 3 things I am still not sure about:
- [math.isclose](https://docs.python.org/3/library/math.html#math.isclose) instead of `==` for Floats.
- pretty-printing of Floats
- Performance of django templates is not ideal. Is there any visible performance penalty for this template expansion?
Other than that I am totally for.

MasloMaslane: - `result_percentage` is now a `Fraction`, so there are no rounding errors
- added
- for ~60 tests each with comment and result percentage, the templates render less than 0.01s slower

twalen (approved):

PR #314, opened 2024-01-06
Fixed string truncation for TestReport's comment

metenn:
Django's `Truncator(text).chars(n)` seems to [ignore all unicode combining characters](https://github.com/django/django/blob/53fc6ac64976a7693d2272050a03b71e56b16578/django/utils/text.py#L125-L128), resulting in incorrect string truncations.    
For example,
```python
string = 'After\u20e1 trunc, this should be 20 characters...'
temp = Truncator(string).chars(20)
print(len(temp))
```
this outputs:
```
21
```
If we were to remove the unicode symbol, it would correctly output 20. Whether this is truly intended or not eludes me. 

When trying to put a string truncated in such manner into a CharField, it fails to pass the MaxLengthValidator check, as [it internally uses len() for its check](https://github.com/django/django/blob/53fc6ac64976a7693d2272050a03b71e56b16578/django/core/validators.py#L472-L473).

Since Python 3, strings respect Unicode by default, so slices seem to work fine for truncation.

### DIFF
--- a/oioioi/programs/handlers.py
+++ b/oioioi/programs/handlers.py
@@ -559,9 +559,8 @@ def make_report(env, kind='NORMAL', save_scores=True, **kwargs):
         comment = result.get('result_string', '')
         if comment.lower() in ['ok', 'time limit exceeded']:  # Annoying
             comment = ''
-        test_report.comment = Truncator(comment).chars(
-            TestReport._meta.get_field('comment').max_length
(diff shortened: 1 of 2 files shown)


MasloMaslane (approved):

MasloMaslane:
This change was tested on sio2.mimuw and works

PR #316, opened 2024-01-10
Update Celery and SQLAlchemy versions

dawidratynski:


twalen (approved):
Looks good to me. But I will also ask @DietPawel if we are ready to merge it.

DietPawel (approved):
I don't think there is much to this change. I believe we use no celery backend in prod and the `sharded_task` is like a rename.

PR #320, opened 2024-01-13
Let contest controllers override settings.COMPLAINTS_EMAIL

### LINKED ISSUE #319: Customizing complaints email per contest
Currently there is only a default email for complaints for whole oioioi. I think there should be an option to override it per contest.

DietPawel:
Fixes #319

### DIFF
--- a/oioioi/complaints/views.py
+++ b/oioioi/complaints/views.py
@@ -35,6 +35,13 @@ def can_make_complaint(request):
         return False
 
 
+def get_complaints_email(request):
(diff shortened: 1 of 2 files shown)


twalen (approved):
it would be nice to add some unit tests but you can merge it right now

DietPawel:
Thanks for pointing it out. I was speaking with Mateusz about it. The current test execute through to override procedure due to implementation. I think that for the long run the test should be refactored not to rely on the OIContestController, as deleting methods from such important class seemed a little dirty to me.

PR #323, opened 2024-01-24
Pin package constraints

### LINKED ISSUE #292: Package constraints in `setup.py` should be stricter
With current constraints, during Szkopuł deployment pip resolves constraints for more than an hour. Pip usually chooses newest versions matching constraints, so making them stricter shouldn't create any issues

dawidratynski:
Add a way to automatically pin the newest compatible package versions using pip-compile based on already present package requirements. This significantly speeds up deployment time.

This approach could potentially be better in the long run, as bumping package versions won't require manually going through all packages to change requirements, which could be an issue if only very strict requirements are present.

Closes #292


A-dead-pixel:
Can't we somehow keep the not-pinned requirements in setup.py?
`pip-compile` supports generating from that.

Dockerfile
```diff
@@ -59,6 +59,7 @@ RUN pip3 install --user bsddb3==6.2.7
 
 WORKDIR /sio2/oioioi
 
+COPY --chown=oioioi:oioioi setup.py requirements_pinned.txt ./
 COPY --chown=oioioi:oioioi setup.py requirements.txt ./
```
A-dead-pixel, line 63: setup.py is copied twice here. Imo all 3 files can be copied in one go.

dawidratynski: Fixed

dawidratynski:
Moved not-pinned requirements back to setup.py and recompiled requirements.

Iteron-dev:
In the current state of oioioi, pinning package constraints using `requirements_pinned.txt` file (where package versions are specified) instead of regular `requirements.txt` doesn't make much of a difference:

- with `requirements_pinned`: `Average execution time: 44.103677 seconds over 10 runs.`

- the "old way": `Average execution time: 43.944880 seconds over 10 runs.`

These values represent the time it takes to install packages in a Docker container (not the entire build process).
I ran these tests locally – the results might not be highly reliable, but even so, the difference is not satisfactory.

So, it’s unlikely that this approach will speed up package installation - the potential improvement might come from better caching and switching to a faster alternative to pip, such as [uv](https://github.com/astral-sh/uv)

Given this, I think that we can close this PR 🫤

PR #334, opened 2024-02-24
Hints for common submission errors

### LINKED ISSUE #326: Explanation of runtime errors
Each year we get a lot of questions "what means that the program exited with code X?"

I would like to see an exact explanation why it returned with code X. For example:
> The code had a runtime error. It returned with code X, which in general means that it dereferenced unused memory, for example because of out-of-box indexing of an array, (...).

I believe there should be a mapping in linux between returned code and unix name of signal.

It would be also helpful to explain that the runtime of code could differ between different environments and running under "test runs" can be helpful.

If the explanation is too long, we could hide it under a spoiler.

(it is a loose example)

Ideally, we would want it to be written and merged into SIO2 before this year's second stage of OI (13.02.2024). Would it be possible @twalen ?

metenn:
This adds hints/explanations for common errors than can show up when submitting to sio2.

Resolves #326

### DIFF
--- a/oioioi/contests/templates/contests/submission.html
+++ b/oioioi/contests/templates/contests/submission.html
@@ -35,7 +35,7 @@
                 {{ report }}
                 <div class="alert alert-info">
                     {% if submission.status == 'SE' %}
-                        <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-info-circle-fill" viewBox="0 0 16 16"><path d="M8 16A8 8 0 1 0 8 0a8 8 0 0 0 0 16m.93-9.412-1 4.705c-.07.34.029.533.304.533.194 0 .487-.07.686-.246l-.088.416c-.287.346-.92.598-1.465.598-.703 0-1.002-.422-.808-1.319l.738-3.468c.064-.293.006-.399-.287-.47l-.451-.081.082-.381 2.29-.287zM8 5.5a1 1 0 1 1 0-2 1 1 0 0 1 0 2"></path></svg>
+                        <i class="fa-solid fa-circle-info"></i>
                         {% trans "Something went wrong on our side when judging your submission... Try again later or contact the administration if this issue persists." %}
                     {% endif %}
                 </div>
--- a/oioioi/contests/templates/contests/submission.html
+++ b/oioioi/contests/templates/contests/submission.html
@@ -31,8 +31,14 @@
     {{ header }}
     <div class="row">
         {% for report in reports %}
-            <div class="{% if submission.status != 'CE' and  submission.status != 'SE' %}col-xl-6{% endif %}">
+            <div class="{% if submission.status != 'CE' and submission.status != 'SE' %}col-xl-6{% endif %}">
                 {{ report }}
+                <div class="alert alert-info">
+                    {% if submission.status == 'SE' %}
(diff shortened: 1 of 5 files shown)


oioioi/programs/templates/programs/report-signal-hint.html
```diff
@@ -0,0 +1,79 @@
...
+                {% endblocktranslate %}
+            </span>
+        {% endif %}
+        <br>
+    {% endfor %}
+    <hr>
+    {% blocktranslate %}
+        Program execution can differ depending on the operating system used and the execution environment.
+        Because of this, we recommend using our Test runs.
+    {% endblocktranslate %}
```
A-dead-pixel, line 78: Many contests (or interactive tasks) don't have test runs, so I would suggest adding "(if available)" at the end.
Also, maybe using Linux could be recommended here too?

metenn: Good point, I'll remove the 2nd line for now.
As for recommending Linux, ideally maybe this note could link to the given contest's "technical rules", [like we have for OI](https://oi.edu.pl/static/attachment/20240129/31oi_etap2_ustalenia_techniczne.pdf), but I think OI/OIJ are the only contests that explicitly declare that, and contestants should be fully aware of them before they could ever reach this message.

oioioi/contests/templates/contests/submission.html
```diff
@@ -31,8 +31,14 @@
     {{ header }}
     <div class="row">
         {% for report in reports %}
-            <div class="{% if submission.status != 'CE' and  submission.status != 'SE' %}col-xl-6{% endif %}">
+            <div class="{% if submission.status != 'CE' and submission.status != 'SE' %}col-xl-6{% endif %}">
                 {{ report }}
+                <div class="alert alert-info">
+                    {% if submission.status == 'SE' %}
+                        <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-info-circle-fill" viewBox="0 0 16 16"><path d="M8 16A8 8 0 1 0 8 0a8 8 0 0 0 0 16m.93-9.412-1 4.705c-.07.34.029.533.304.533.194 0 .487-.07.686-.246l-.088.416c-.287.346-.92.598-1.465.598-.703 0-1.002-.422-.808-1.319l.738-3.468c.064-.293.006-.399-.287-.47l-.451-.081.082-.381 2.29-.287zM8 5.5a1 1 0 1 1 0-2 1 1 0 0 1 0 2"></path></svg>
```
twalen, line 38: can we use something like: https://icons.getbootstrap.com/icons/info-circle/

oioioi/programs/templates/programs/report-signal-hint.html
```diff
@@ -0,0 +1,78 @@
+{% load i18n %}
+
+<div class="alert alert-info">
+    <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-info-circle-fill" viewBox="0 0 16 16"><path d="M8 16A8 8 0 1 0 8 0a8 8 0 0 0 0 16m.93-9.412-1 4.705c-.07.34.029.533.304.533.194 0 .487-.07.686-.246l-.088.416c-.287.346-.92.598-1.465.598-.703 0-1.002-.422-.808-1.319l.738-3.468c.064-.293.006-.399-.287-.47l-.451-.081.082-.381 2.29-.287zM8 5.5a1 1 0 1 1 0-2 1 1 0 0 1 0 2"></path></svg>
+    <br>
+    {% for signal in signals_to_explain %}
+        {% if signal == 6 %}
+            {% blocktrans %}
+                <b>Most common causes of the SIGABRT signal (signal 6):</b>
+                <br>
```
twalen, line 10: is there any reason not use <UL>?

oioioi/contests/templates/contests/submission.html
metenn: I'll replace it with an icon from FontAwesome, which sio2 already has installed/support for.
https://fontawesome.com/icons/circle-info?f=classic&s=solid

oioioi/programs/templates/programs/report-signal-hint.html
metenn: You're right. It looks better, too. I fixed this in the newest commit.

twalen (approved):

PR #335, opened 2024-02-28
Reduce CSS bundle

fmkra:
The main cause of CSS bundle being so big is that `scss/style` is imported several times in different files and isn't deduplicated by django. Fixed by replacing that import with one that includes only specific, necessary file. This reduced CSS bundle from around 1MB to 350kB.


twalen:
Is this PR ready for a review?

twalen (approved):
OK

PR #336, opened 2024-02-28
Display contest rules

NadiaHoffmann:
Added the subpage with various information about the contest, such as its type, beginning and end date and time, number of rounds, submission limits etc. More specific info about the contest scoring could be added.


oioioi/contests/views.py
```diff
@@ -102,6 +106,49 @@ def get_contest_permissions(request, response):
     return response
 
 
+@menu_registry.register_decorator(
+    _("Rules"), lambda request: reverse('contest_rules'), order=100
```
matihope, line 110: There are other entries that have `order=100`. I'd suggest changing it to something like `90`, because now the real order is somewhat nondeterministic.

oioioi/contests/views.py
```diff
@@ -102,6 +106,49 @@ def get_contest_permissions(request, response):
...
+        contest_start_date = None
+
+    if (contest_dates[1] != -1):
+        contest_end_date = contest_dates[1]
+    else :
+        contest_end_date = None
+
+    if (len(submission_limit) == 1 
+        and (submission_limit[0] == None 
+        or submission_limit[0] == 0)):
```
matihope, line 132: Style - redundant braces.

oioioi/contests/utils.py
```diff
@@ -247,6 +256,96 @@ def visible_rounds(request):
     return [r for r in queryset if controller.can_see_round(request, r)]
 
 
+@request_cached
+def get_number_of_rounds(request):
+    """Returns the number of rounds in the current contest.
+    """ 
+    return Round.objects.filter(contest=request.contest).count()
```
matihope, line 263: this can be replaced with `get_visible_rounds` if I am correct.

oioioi/contests/utils.py
```diff
@@ -247,6 +256,96 @@ def visible_rounds(request):
...
+    ends = [
+        rt.get_end()
+        for rt in rtimes.values()
+    ]
+    starts = [
+        rt.get_start()
+        for rt in rtimes.values()
+    ]
+
+    if (not starts or starts.count(None)):
```
matihope, line 284: Style - redundant braces.

oioioi/contests/utils.py
```diff
@@ -247,6 +256,96 @@ def visible_rounds(request):
...
+        min_start = -1
+    else:
+        min_start = min(starts)
+
+    if (not ends or ends.count(None)):
```
matihope, line 289: Style - redundant braces.

oioioi/contests/utils.py
```diff
@@ -247,6 +256,96 @@ def visible_rounds(request):
...
+    for r in rtimes.keys():
+        results_date = rtimes[r].results_date()
+        public_results_date = rtimes[r].public_results_date()
+ 
+        if (results_date is None or results_date <= request.timestamp):
+            results_visibility = 'immediately'
+        else:
+            results_visibility = f'after {results_date.strftime("%Y-%m-%d %H:%M:%S")}'
+
+        if (public_results_date is None or public_results_date <= request.timestamp):
```
matihope, line 335: Style - redundant braces.

oioioi/contests/utils.py
```diff
@@ -247,6 +256,96 @@ def visible_rounds(request):
...
+        rtimes = rounds_times(request, request.contest)
+    else:
+        rtimes = generic_rounds_times(None, request.contest)
+
+    dates = list()
+    for r in rtimes.keys():
+        results_date = rtimes[r].results_date()
+        public_results_date = rtimes[r].public_results_date()
+ 
+        if (results_date is None or results_date <= request.timestamp):
```
matihope, line 330: Style - redundant braces.

oioioi/contests/utils.py
```diff
@@ -247,6 +256,96 @@ def visible_rounds(request):
...
+
+
+def get_contest_dates(request):
+    """Returns the end_date of the latest round and the start_date
+    of the earliest round.
+    """ 
+    if isinstance(request, HttpRequest):
+        rtimes = rounds_times(request, request.contest)
+    else:
+        rtimes = generic_rounds_times(None, request.contest)
```
matihope, line 273: Can you explain the logic here? To me, the other branch never happens.

oioioi/contests/utils.py
```diff
@@ -247,6 +256,96 @@ def visible_rounds(request):
...
+
+    if (not starts or starts.count(None)):
+        min_start = -1
+    else:
+        min_start = min(starts)
+
+    if (not ends or ends.count(None)):
+        max_end = -1
+    else:
+        max_end = max(ends)
```
matihope, line 292: I'd say
```py
if starts and None not in starts:
    min_start = min(starts)
else:
    min_start = -1
```
To me, this is more self explanatory.

oioioi/contests/utils.py
```diff
@@ -247,6 +256,96 @@ def visible_rounds(request):
...
+def get_problems_sumbmission_limit(request):
+    """Returns the list of distinct submission limits in the current contest.
+    """
+    controller = request.contest.controller
+    queryset = (
+        ProblemInstance.objects.filter(contest=request.contest)
+        .prefetch_related('round')
+    )
+
+    if (queryset is None):
```
matihope, line 307: Style - redundant braces.

oioioi/contests/utils.py
```diff
@@ -247,6 +256,96 @@ def visible_rounds(request):
...
+    return sorted(limits)
+
+
+@request_cached
+def get_results_visibility(request):
+    """Returns the results visibility for each round in the contest"""
+    if isinstance(request, HttpRequest):
+        rtimes = rounds_times(request, request.contest)
+    else:
+        rtimes = generic_rounds_times(None, request.contest)
```
matihope, line 323: Again, the other branch is never? chosen.

oioioi/contests/views.py
```diff
@@ -102,6 +106,49 @@ def get_contest_permissions(request, response):
...
+)
+@enforce_condition(contest_exists & can_enter_contest)
+def contest_rules_view(request):
+    no_of_rounds = get_number_of_rounds(request)
+    contest_dates = get_contest_dates(request)
+    submission_limit = get_problems_sumbmission_limit(request)
+    results_visibility = get_results_visibility(request)
+    scoring_type = request.contest.controller.description
+
+    if (contest_dates[0] != -1):
```
matihope, line 120: Style - redundant braces.

oioioi/contests/views.py
```diff
@@ -102,6 +106,49 @@ def get_contest_permissions(request, response):
...
+        contest_start_date = contest_dates[0]
+    else :
+        contest_start_date = None
+
+    if (contest_dates[1] != -1):
```
matihope, line 125: Style - redundant braces.

oioioi/contests/templates/contests/contest_rules.html
```diff
@@ -0,0 +1,54 @@
...
+{% load i18n %}
+{% load pagination_tags %}
+
+{% block title %}{% trans "Rules" %}{% endblock %}
+
+{% block main-content %}
+<h1>{% trans "Rules" %}</h1>
+<ul class="list-group list-group-light">
+    {% if scoring_type %}
+        {% blocktrans %}
```
matihope, line 11: The recent syntax is `blocktranslate`. Does the text appear in the translation files?

oioioi/contests/templates/contests/contest_rules.html
```diff
@@ -0,0 +1,54 @@
...
+            <li class="list-group-item">The scoring and contest type is "{{ scoring_type }}".</li>
```
matihope, line 12: HTML blocks used inside `blocktranslate` are omitted? If so, that's fine.

easy_toolbox.py
```diff
@@ -36,8 +36,10 @@
      "{exec} web python manage.py loaddata ../oioioi/oioioi_cypress/cypress/fixtures/admin_admin.json"),
     ("test", "Run unit tests.", "{exec} web ../oioioi/test.sh"),
     ("test-slow", "Run unit tests. (--runslow)", "{exec} web ../oioioi/test.sh --runslow"),
-    ("test-abc", "Run specific test file. (edit the toolbox)",
-     "{exec} web ../oioioi/test.sh -v oioioi/teachers/tests.py"),
+    ("test-abc", "Run specific test file. The path is hardcoded in the toolbox.", 
+     "{exec} web ../oioioi/test.sh -v oioioi/contests/tests/tests.py::TestRulesVisibility"),
+    ("test-specific", "Run specific test file. You will be promted to enter the path to the test file, starting from oioioi/[...].",
```
matihope, line 41: Maybe add a support for an argument.

matihope (changes requested):
Great work. Minor comments, but pleaaase shorten the tests, because they are mind-boggling.

oioioi/contests/templates/contests/contest_rules.html
MasloMaslane: `blocktrans` and `trans` are still aliases for `blocktranslate` and `translate`

oioioi/contests/views.py
```diff
@@ -102,6 +106,49 @@ def get_contest_permissions(request, response):
...
+@menu_registry.register_decorator(
+    _("Rules"), lambda request: reverse('contest_rules'), order=100
+)
+@enforce_condition(contest_exists & can_enter_contest)
+def contest_rules_view(request):
+    no_of_rounds = get_number_of_rounds(request)
+    contest_dates = get_contest_dates(request)
+    submission_limit = get_problems_sumbmission_limit(request)
+    results_visibility = get_results_visibility(request)
+    scoring_type = request.contest.controller.description
```
MasloMaslane, line 118: Personally, I’m not sure if `description` describes well the scoring type. For example `Polish Olympiad in Informatics - Onsite` doesn’t describe how the submission are scored and on sio2.mimuw we have even less meaningful descriptions.

Maybe a solution would be to have a field `scoring_description` on the contest controller and if this field is not present, the scoring type can be hidden.

MasloMaslane (changes requested):
I would also like to have an option to hide the rules, for example with a function on the contest controller

oioioi/contests/utils.py
```diff
@@ -247,6 +256,96 @@ def visible_rounds(request):
...
+    )
+
+    if (queryset is None):
+        return None
+    
+    limits = set()
+    for p in queryset:
+        limits.add(controller.get_submissions_limit(request, p))
+
+    return sorted(limits)
```
A-dead-pixel, line 314: I think that for contests without problems the contest's default submissions limit should be returned here.

oioioi/contests/utils.py
```diff
@@ -247,6 +256,96 @@ def visible_rounds(request):
...
+ 
+        if (results_date is None or results_date <= request.timestamp):
+            results_visibility = 'immediately'
+        else:
+            results_visibility = f'after {results_date.strftime("%Y-%m-%d %H:%M:%S")}'
+
+        if (public_results_date is None or public_results_date <= request.timestamp):
+            public_results_visibility = 'immediately'
+        else:
+            public_results_visibility = f'after {public_results_date.strftime("%Y-%m-%d %H:%M:%S")}'
```
A-dead-pixel, line 338: These strings should be translated.

A-dead-pixel (changes requested):

oioioi/contests/templates/contests/contest_rules.html
NadiaHoffmann: My bad, the HTML blocks aren't omitted - thank you for pointing this one out!

oioioi/contests/utils.py
NadiaHoffmann: visible_rounds() returns only rounds that a user can see, and by deafult it does not show the rounds that will start in the future. My train of thought was to inform the user how many rounds are in the contest including ones in the future. However, I may change it if counting only visible rounds is a better option (as seeing that contest has more rounds than the visible ones might be confusing for some users).

oioioi/contests/utils.py
NadiaHoffmann: You're right, corrected.

oioioi/contests/utils.py
NadiaHoffmann: c habit... fixed

oioioi/contests/utils.py
NadiaHoffmann: You're right, this seems much better. Fixed!

oioioi/contests/utils.py
NadiaHoffmann: Good idea, changed.

oioioi/contests/utils.py
NadiaHoffmann: As far as I understand, in the recent commit I fixed it so that they're translated in a template - but in case I'm wrong, please let me know!

oioioi/contests/views.py
NadiaHoffmann: Good idea, implemented it in such a way. Whom I could contact so that I knwo what to write in `scoring_description` fields in different contest controllers?

NadiaHoffmann:
> I would also like to have an option to hide the rules, for example with a function on the contest controller

I just want to make sure, what do you mean by "hiding the rules" exactly? Giving a contest admin an option in contest settings to hide the rules so that the "Rules" tab is not displayed at all and the page is not accessible, or something else?

MasloMaslane:
Yes, an option to disable the "Rules" page in contest settings would be perfect

oioioi/contests/views.py
MasloMaslane: I'll write the descriptions now

MasloMaslane: ProgrammingContestController
The submissions are scored on a set of groups of test cases. Each group is worth a certain number of points.
The score is a sum of the scores of all groups. The ranking is determined by the total score.
The full scoring is available after the results date for the round.

ACMContestController, ACMOpenContestController:
The solutions are judged on real-time.
The submission is correct if it passes all the test cases.
Participants are ranked by the number of solved problems. 
In case of a tie, the times of first correct submissions are summed up and a penalty of 20 minutes is added for each incorrect submission.
The lower the total time, the higher the rank.
Compilation errors and system errors are not considered as an incorrect submission.
The ranking is frozen 60 minutes before the end of the round.

AMPPZContestController:
same as ACMContestController, but the ranking is frozen 15 minutes before the end of the trial rounds and 60 minutes before the end of the normal rounds.

MPContestController:
The submissions are scored from 0 to 100 points. The participant can submit to finished rounds,
but a multiplier is applied to the score of such submissions.

OIContestController, OIOnsiteContestController:
The solutions are judged with sio2jail. They can be scored from 0 to 100 points.
If the submission runs for longer than half of the time limit, the points for this test are linearly decreased to 0.
The score for a group of test cases is the minimum score for any of the test cases.
The ranking is determined by the total score.
Until the end of the contest, participants can only see scoring of their submissions on example test cases.
Full scoring is available after the end of the contest.

OIFinalOnsiteContestController:
Same as OIContestController, but participants can reveal the full scoring of their submissions during the contest.

BOIOnsiteContestController, BOIOnlineContestController: not used

PAContestController:
The submissions are judged on real-time. All problems have 10 test groups, each worth 1 point.
If any of the tests in a group fails, the group is worth 0 points.
The ranking is determined by the total score.
The full scoring is available after the end of the round.
The ranking is determined by the total score and number of 10-score submissions, 9-score, 8...

PAFinalsContestController:
same as AMPPZContestController

oioioi/contests/utils.py
matihope: Okay, that's a problem where either solution could be chosen, because one or the other can be better depending on a situation. I am not in a position to make a final decision here, but I'd say all rounds should be counted, because it doesn't really reveal any crucial information about contest's problems.

matihope: Also, since hiding the rules will be made possible (or already has been), then it is always possible to hide this information if needed.

oioioi/contests/utils.py
A-dead-pixel: Unfortunately, `{{ blocktrans with value as var }}` doesn't translate `value`, it only conditionalizes the string to translate based on the contents of `value`, which in this basic case means just inserting the `value` string in appropriate places.
I myself wasn't sure if that was the case, but running `cd /sio2/oioioi && /sio2/deployment/manage.py makemessages -a` in the web container shows no trace of these strings in `.po` files.

oioioi/contests/views.py
```diff
@@ -102,6 +107,70 @@ def get_contest_permissions(request, response):
...
+        contest_start_date = contest_dates[0]
+    else :
+        contest_start_date = None
+
+    if contest_dates[1] != -1:
+        contest_end_date = contest_dates[1]
+    else :
+        contest_end_date = None
+
+    submission_limit = get_problems_sumbmission_limit(request)
```
twalen, line 152: it does not work for admin (it always show -> no limit)

oioioi/contests/utils.py
```diff
@@ -250,6 +260,87 @@ def visible_rounds(request, no_admin=False):
...
+        max_end = max(ends)  
+    else:  
+        max_end = -1  
+        
+
+    return min_start, max_end
+
+
+@request_cached
+def get_problems_sumbmission_limit(request):
```
twalen, line 308: I encourage to move all logic here and leave just rendering in view / template.
Also I would recommend to reduce the information just to:
- no limit
- X submission per problem (if there is only single limit for all tasks)
- X to Y submissions depending on the problem (if there are multiple different values)

MasloMaslane (approved):

matihope (approved):

Issue #354, opened 2024-03-24
Submission error hints are not appearing for testruns


A-dead-pixel:
When I tested it locally, only an empty box appeared:

![image](https://github.com/sio2project/oioioi/assets/82606748/b03ca59e-01ac-4eb4-abc4-af6d7d6ceadd)

metenn:
Error hints not showing up for test runs was something I overlooked and thus they shouldn't show up at all, so the empty box is probably from something else?

A-dead-pixel:
This isn't completed yet, as only the suspicious blue box has been fixed (by #355).
Testruns still don't have error hints.

metenn:
I'm on it!

twalen:
Why in the screenshot you have a comment:
"program exited due to signal 6"?

Both oioioi code & sioworkers code expect prefix:
"process exited due to signal "
(see https://github.com/sio2project/oioioi/blob/master/oioioi/programs/controllers.py#L740)

A-dead-pixel:
Two quick `rg`s show that [sioworkers](https://github.com/sio2project/sioworkers/blob/9137c6a3726dd6de422dc79ed16319d782ec78f4/sio/workers/executors.py#L419-L426) are the culprit and `program exited due to...` also shows up in oioioi tests.

PR #355, opened 2024-03-24
SE error hint fix

metenn:
Fixes the empty blue box error in #354 mistakenly introduced in #334

The div and if statement for checking if a submission had a system error were misplaced.


twalen (approved):
ok

PR #360, opened 2024-04-06
Fix sections underlines in README.rst

afurtak:
A few section underlines were too short and they werent rendered correctly.
Link to the specification:
https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#section


twalen:
What happened?

Issue #352, opened 2024-03-20
Migration 0002 in OI app breaks clearing database


dawidratynski:
Migration 0002 in OI app causes db clearing to not work, by breaking the `python manage.py flush` command (and by extension: `./easytoolbox.py flush-db`), which is needed for running cypress tests locally.

https://github.com/sio2project/oioioi/blob/c3c399b02312dc1420aa43b474b9c01d10e9463e/oioioi/oi/migrations/0002_auto_20160412_1720.py

This migration removes Region and OIOnsiteRegistration models from the OI app, but does not delete their tables from the database (`migrations.SeparateDatabaseAndState`). Said tables contain non-nullable foreign key fields to other tables, causing an error when trying to clear the db, as Django only clears tables related to existing objects.

Depending on whether deleting these tables is acceptable, I found two possible solutions:

1. If we can't delete the tables, we can make the foreign key fields nullable:
```
database_operations = [
        migrations.AlterField(
            model_name='oionsiteregistration',
            name='region',
            field=models.IntegerField(verbose_name='region', null=True),
        ),
        migrations.AlterField(
            model_name='oionsiteregistration',
            name='participant',
            field=oioioi.participants.fields.OneToOneBothHandsCascadingParticipantField(related_name='oi_oionsiteregistration', to='participants.Participant', on_delete=models.CASCADE, null=True),
        ),
        migrations.AlterField(
            model_name='region',
            name='contest',
            field=models.ForeignKey(to='contests.Contest', on_delete=models.CASCADE, null=True),
        )
    ]

    state_operations = [
        migrations.DeleteModel('oionsiteregistration'),
        migrations.DeleteModel('region'),
    ]

    operations = [
        migrations.SeparateDatabaseAndState(
            database_operations=database_operations,
            state_operations=state_operations)
    ]
```

2. If we can delete said tables:
```
operations = [
        migrations.DeleteModel(
            name='OIOnsiteRegistration',
        ),
        migrations.DeleteModel(
            name='Region',
        ),
    ]
```

Both solutions fix the issue and pass all tests (as of 19.03.2024 / commit 056caf6).

A-dead-pixel:
This is probably caused by f55acf60fdc39f0bde9bc66762ea4a8ad1dfea92.

dawidratynski:
Created branches with implemented solutions for testing:

With deleting tables: 
https://github.com/dawidratynski/oioioi/tree/fix-oi-migration-v1

Without deleting tables:
[...]

PR #362, opened 2024-04-10
Upgrade Cypress and fix e2e tests

dawidratynski:
1. Upgrade Cypress (to version ^13.7.2) and its dependencies.
2. Change preexisting e2e tests to be idempotent (this fixes a bug that caused some tests to fail if ran multiple times) and work without flushing db. This allows Cypress tests to be run locally, despite issue #352.


oioioi_cypress/cypress/e2e/user/register.cy.ts
```diff
@@ -1,14 +1,17 @@
...
     before(() => {
         cy.visit("/");
         cy.hideDjangoToolbar();
-        cy.enLang()
+        cy.enLang();
     });
 
     it("Register new user", () => {
         cy.fixture("credentials").then((data) => {
+            data.user.username = getUniqueUsername();
```
twalen, line 14: Could you add a comment why unique user is needed?

twalen (changes requested):
The PR looks OK, just add some comment such that the getUniqueUser will not be accidentally refactored at some point.

twalen (approved):

PR #363, opened 2024-04-11
Consistent new message formatting for new questions

metenn:
ATM new questions ignore newlines, which is inconsistent and can make the incoming new message hard to read.

Note: I'm not sure if this violates BEM naming conventions, [some claim it doesn't](https://en.bem.info/forum/43/). If it does I'm not against nesting it in `.oioioi-message`, since that class when by itself currently has no styling: https://github.com/sio2project/oioioi/blob/0fcee5ec26341c8820c71d7ea4f104f00e6d7e6f/oioioi/questions/static/common/messages.scss#L3-L41


twalen (approved):
OK

PR #366, opened 2024-04-17
Fix oi migration

dawidratynski:
Old migration in OI app removed two db tables from Django models without deleting them from the db. As they contained non-nullable foreign keys, the db couldn't be flushed. 

This pull request changes said columns to be nullable, fixing the issue without deleting any data from the db.

Full description in issue #352 description. 

Closes issue #352


twalen (approved):
ok

PR #367, opened 2024-04-17
Fix unnecesary hardcoded path

### LINKED ISSUE #340: Fix hardcorded path
https://github.com/sio2project/oioioi/blob/master/oioioi/deployment/supervisord.conf.template#L94

dawidratynski:
Remove unnecessary hardcoded path in deployment template.

Closes #340


twalen (approved):
OK

PR #370, opened 2024-04-21
Fix search and filtering in SubmissionAdmin

A-dead-pixel:
This fixes a bug where the submission queryset contained objects matching a search term *or* other filters, not *and*.
It was caused by using `self.get_queryset(request)` instead of the `queryset` passed in args, which sometimes bypassed user-selected filters.

Steps to reproduce:
1. Add two problems with different names to a contest, let's say "abc" and "tst"
2. Add ProblemName objects to "abc", e.g. "aaa"
3. Go to Contest Administration -> Submissions, filter "by problem" for "tst"
4. Search for "aaa"
5. See submissions from "abc", despite the filter for "tst"

### DIFF
--- a/oioioi/contests/admin.py
+++ b/oioioi/contests/admin.py
@@ -685,6 +685,7 @@ class SubmissionAdmin(admin.ModelAdmin):
         'user__last_name',
         'problem_instance__problem__legacy_name',
         'problem_instance__short_name',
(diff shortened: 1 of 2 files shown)


A-dead-pixel:
Thanks to @metenn for a neater implementation!

twalen (approved):
OK

PR #371, opened 2024-04-21
Add a test for missing migrations and add them

### LINKED ISSUE #256: We should fail PR check if makemigrations proposes any changes
Currently it is easy to suggest code, which does not include all required migrations from `makemigrations`.

A-dead-pixel:
Closes #256.
Unfortunately, this only works if pytest is ran with `--migrations`, but github workflows use this flag.


twalen (approved):
Looks good. I will modify github action.

A-dead-pixel:
The issue seems to be tracked here: https://github.com/microsoft/linux-package-repositories/issues/130.

PR #373, opened 2024-04-24
Api improvements

matihope:
This is a draft PR showcasing current functionality implemented in the API. In terms of problems it allows for:

* listing contests
* listing problems in a contest
* getting details about a problem in a contest
* submitting a solution to a problem in a contest (This was already implemented)

Currently, yet to be implemented is querying a submission list, a ranking, writing tests and request limitation.

___

# VSCode Extension

This issue is being worked on in parallel with a VSCode extension, that already allows to submit solutions right from the editor. What a time to be alive!


MasloMaslane:
I have a concern: if there is an onsite contest with ip authentication enabled is it possible to submit as someone else with their api key?

oioioi/contests/api.py
```diff
@@ -1,23 +1,246 @@
...
 
 
 class CanEnterContest(permissions.BasePermission):
     def has_object_permission(self, request, view, obj):
         return can_enter_contest(request)
 
 
+class UnsafeApiAllowed(permissions.BasePermission):
+    def has_object_permission(self, request, view, obj):
+        # TODO: Is that ok?
```
twalen, line 53: remove comment + add
```
return not any('IpDnsAuthMiddleware' in x for x in MIDDLEWARE)
```

oioioi/contests/utils.py
```diff
@@ -637,3 +635,47 @@ def has_view_permission(self, request, obj=None):
             return True
 
     return ArchivedInlineWrapper
+
+
+def get_problem_statements(request, controller, problem_instances):
```
twalen, line 640: it would be a nightmare to maintain this, what about returning list of dicts?

twalen (changes requested):
Please handle the comments.

oioioi/contests/urls.py
```diff
@@ -150,7 +150,11 @@ def glob_namespaced_patterns(namespace):
         name='user_info_redirect',
     ),
     re_path(r'^admin/', admin.contest_site.urls),
-    re_path(r'^archive/confirm$', views.confirm_archive_contest, name='confirm_archive_contest'),
```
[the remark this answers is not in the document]
matihope: It's not a deletion, only a format. But I'll revert it.

oioioi/contests/utils.py
matihope: Can you elaborate please?

twalen: In this function you return tuple with very specific order of tuple-elements. This ordering is repeated in two distinct places and the code would fail if they will be out-of-sync.

I suggest to have a get_problem_statements(..) that returns a list of dictionaries i.e.
```
[
  {"problem_instance": pi1, "statement_visible": True, "round_end_time": datetime(...), ...},
  {"problem_instance": pi2, "statement_visible": False, "round_end_time": datetime(...), ...},
   ...
]
```

With this change you still need to maintain sync between different fragments of code, but at least you will get an error in case of mispelled key.

twalen: Other then this change I'm happy to merge the PR.

twalen (approved):
Please resolve merge conflicts and we can merge it.

twalen (approved):

PR #374, opened 2024-05-05
Hide problems field from OriginTag's and OriginInfoValue's admin view

MasloMaslane:
When the oioioi instance has a lot of problems, the views load a very long time (on szkopuł they take more than a minute and they timeout). I don't think many people use origin tags and probably even less use the problem field in the admin views, so I believe this is safe to hide.


twalen (approved):
Looks good.

PR #375, opened 2024-05-07
Add start round date 

PPiotrek12:


NadiaHoffmann (dismissed):
Smart solution with template tag, looks good to me :)

oioioi/base/templatetags/format_data_range.py
```diff
@@ -0,0 +1,20 @@
+from django import template
+
+register = template.Library()
+
+@register.simple_tag
+def format_data_range(start, end):
```
twalen, line 6: Currently the %b is returning month name according to server locale (usually EN).
So even with polish language selected in user profile it would return sth like:
```(04 Oct 2023, 12:03 - 04 Oct 2099, 14:03)```

I recommend to use sth like:
```
from django.utils import formats

...
        start_str = formats.date_format(start, "j E Y, H:i")
```

File: oioioi/contests/tests/tests.py [hunk 2]
```diff
@@ -941,6 +941,17 @@ def test_mixin_past_rounds_hidden_during_prep_time(self):
         response = self.client.get(reverse('select_contest'))
         self.assertEqual(len(response.context['contests']), 1)
 
+    def test_round_dates(self):
```
twalen, line 944: please add also test for polish locale

twalen (changes requested):

PPiotrek12:
Add date translation

twalen (approved):
OK

PR #378, opened 2024-05-15
Interactive problems via I/O

geoff128:


oioioi/deployment/settings.py.template
```diff
@@ -196,6 +196,7 @@ INSTALLED_APPS = (
     # 'oioioi.problemsharing',
     # 'oioioi.usergroups',
     # 'oioioi.usercontests',
+    'oioioi.interactive',
```
MasloMaslane, line 199: I don’t think this should be both in default setting and the template

MasloMaslane (changes requested):

oioioi/sinolpack/package.py
```diff
@@ -142,6 +143,7 @@ def __init__(self, path, original_filename=None):
         self.restrict_html = (
             settings.SINOLPACK_RESTRICT_HTML and not settings.USE_SINOLPACK_MAKEFILES
         )
+        self.task_type = 'standard'
```
DietPawel, line 146: How about some enum class for this?

oioioi/interactive/models.py
```diff
@@ -0,0 +1,19 @@
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+from oioioi.filetracker.fields import FileField
+from oioioi.problems.models import Problem, make_problem_filename
+
+
+class Interactor(models.Model):
+    problem = models.OneToOneField(Problem, on_delete=models.CASCADE)
```
metenn, line 9: Since this is a one-to-one model, perhaps it is better to use [model iheritance](https://docs.djangoproject.com/en/4.0/topics/db/models/#model-inheritance) here?

metenn (approved):

MasloMaslane (approved):

MasloMaslane (approved):

Issue #383, opened 2024-05-22
Round dates & times shown on tasks page are invalid.


twalen:
It looks like PR #375 introduced an error in presenting start/end round date on the task page, i.e.

https://szkopul.edu.pl/c/laboratorium-z-asd-2023/p/

<img width="820" alt="image" src="https://github.com/sio2project/oioioi/assets/7008315/bc05d347-fbc3-44d8-b236-04a66475e0b5">

it should be "17 lutego 23:59"

MasloMaslane:
Completed in #384

PR #384, opened 2024-05-22
Fix to #383 issue

PPiotrek12:


oioioi/base/templatetags/format_data_range.py
```diff
@@ -1,11 +1,15 @@
...
 from django.utils import formats
 
 register = template.Library()
 
-@register.simple_tag
-def format_data_range(start, end):
+@register.simple_tag(takes_context=True)
+def format_data_range(context, start, end):
     if start is not None and end is not None:
+        start = template_localtime(start, context.use_tz)
```
twalen, line 10: if start is not None and end is not None, then both start/end are passed through template_localtime 2 times (in line 10 & in line 21),
I suggest to move initialisation to the beginning of the function, like:
start = template_localtime(start, context.use_tz) is start is not None else None

twalen (changes requested):

twalen (approved):

PR #387, opened 2024-05-28
livedata: fix cache_unless_admin_or_observer

metenn:
result.content is of type "bytes", so wrapping it in str() would create a string that's wrapped in `b'...'`, in turn causing `JSON.parse()` on the client-side scripts for oioioi.liveranking to fail.


twalen (changes requested):
Could you also add a regression test?

twalen (approved):

PR #391, opened 2024-06-05
Fix score badge

geoff128:
On ACM type contests, the badge's color was red even when the task was accepted.

### DIFF
--- a/oioioi/acm/tests.py
+++ b/oioioi/acm/tests.py
@@ -1,3 +1,4 @@
+import bs4
 import re
 from datetime import datetime, timezone  # pylint: disable=E0611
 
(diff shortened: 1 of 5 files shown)


oioioi/acm/tests.py
```diff
@@ -145,3 +146,26 @@ def test_model_solution_submission_view(self):
...
         contest = Contest.objects.get()
         self.assertEqual(contest.controller.get_safe_exec_mode(), 'cpu')
+
+
+class TestACMScores(TestCase):
+    fixtures = ['acm_test_full_contest']
+
+    def _get_badge_for_problem(self, content, problem):
+        soup = bs4.BeautifulSoup(content, 'html.parser')
+        print(content)
```
twalen, line 156: please remove debug print

twalen (changes requested):
looks ok, but please remove debug print

oioioi/acm/tests.py
geoff128: Apparently, I forgot to git add this file... Fixed!

PR #388, opened 2024-05-29
Change attachments order

MasloMaslane:
Closes #351

PR #392, opened 2024-06-05
Attachments in alphabetic order in groups

MasloMaslane:
I changed #388, so that attachments are sorted in alphabetic order in groups

### DIFF
--- a/oioioi/contests/views.py
+++ b/oioioi/contests/views.py
@@ -539,6 +539,7 @@ def contest_files_view(request):
 
     round_file_exists = contest_files.filter(round__isnull=False).exists()
     add_category_field = round_file_exists or problem_files.exists()
+    all_rows = []
     rows = [
         {
             'category': cf.round if cf.round else '',
@@ -553,7 +554,10 @@ def contest_files_view(request):
         }
         for cf in contest_files
     ]
-    rows += [
+    rows.sort(key=itemgetter('name'))
+    all_rows += rows
+
+    rows = [
         {
             'category': pf.problem,
             'name': pf.download_name,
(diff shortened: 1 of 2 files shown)


A-dead-pixel:
The Downloads view in problem sites should also have sorting for `additional_files`. The group ordering seems to already be present.
https://github.com/sio2project/oioioi/blob/a629e06f17647ee5d3ca333bec683e0a60dad295/oioioi/problems/problem_site.py#L136-L139

oioioi/contests/views.py
```diff
@@ -553,7 +554,10 @@ def contest_files_view(request):
         }
         for cf in contest_files
     ]
-    rows += [
+    rows.sort(key=itemgetter('name'))
```
SZonkil, line 557: We could use .sorted to make is cleaner and shorter

oioioi/contests/views.py
```diff
@@ -567,12 +571,15 @@ def contest_files_view(request):
         }
         for pf in problem_files
     ]
-    rows += additional_files
+    rows.sort(key=itemgetter('name'))
```
SZonkil, line 574: Also here

oioioi/contests/views.py
```diff
@@ -567,12 +571,15 @@ def contest_files_view(request):
...
+    all_rows += rows
+    additional_files.sort(key=itemgetter('name'))
```
SZonkil, line 576: And here

SZonkil (changes requested):

SZonkil (approved):
OK

twalen (approved):

PR #400, opened 2024-09-04
su for contest admins

MasloMaslane:
This PR adds an option for contest admins to su to participants in contests they admin. By default this option is disabled and can be enabled by setting `CONTEST_ADMINS_CAN_SU` to `True` in `settings.py`. Also by default, contest admins using su can't make requests other than GET. This can be changed by setting `ALLOW_ONLY_GET_FOR_SU_CONTEST_ADMINS` to `True` in settings.

When a contest admin using su tries to change to another contest or go to an url which is not in the contest (doesn't have `/c/id/...`), they are redirected to the dashboard of the original contest. This unfortunately still allows to go to to sites like editing profile info, because the url starts with `/c/id/`. This is the reason that by default only GET requests are allowed.

This feature is highly sought after by the OI Jury


oioioi/su/views.py
```diff
@@ -38,6 +41,10 @@ def su_view(request, next_page=None, redirect_field_name=REDIRECT_FIELD_NAME):
     user = form.cleaned_data['user']
     if is_under_su(request):
         raise SuspiciousOperation
+    if not is_superuser(request) and not Participant.objects.filter(
+        user=user, contest=request.contest
+    ).exists():
+        raise SuspiciousOperation
```
A-dead-pixel, line 47: Basicadmins shouldn't be allowed to su into contest admins/owners (and contest admins into owners).
The check for a Participant isn't sufficient because a contest admin could add an owner as a participant or the admins could just be participants, which does occur on my deployments.
This should also be addressed in `get_suable_users_view`.

oioioi/su/middleware.py
```diff
@@ -39,6 +51,40 @@ def _process_request(self, request):
...
+                m = contest_re.match(request.path)
+                if m is not None:
+                    contest_id = m.group('c_name')
+                url = resolve(request.path_info)
+                is_su_reset_url = url.url_name == 'su_reset'
+                nonoioioi_namespace = url.namespaces == []
+                for ns in url.namespaces:
+                    if ns != 'contest' and ns != 'noncontest':
+                        nonoioioi_namespace = True
+                        break
```
A-dead-pixel, line 67: If `oioioiadmin` isn't added to BLOCKED_URL_NAMESPACES, then this logic won't redirect in the following situation:
we are `su`ed into a user, who is a contest admin in another contest and we visit some admin view in that contest.

A-dead-pixel (changes requested):

oioioi/su/views.py
```diff
@@ -39,6 +43,23 @@ def su_view(request, next_page=None, redirect_field_name=REDIRECT_FIELD_NAME):
...
         raise SuspiciousOperation
 
+    # Don't allow contest admins to switch to users that are not participants of the contest
+    if not is_superuser(request) and not Participant.objects.filter(
+        user=user,
+        contest=request.contest,
+        user__is_active=True,
+        user__is_superuser=False,
+    ).exists():
+        raise SuspiciousOperation
```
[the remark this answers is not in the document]
MasloMaslane: How should `su`ing in public contests work? I think it shouldn't be allowed?

oioioi/su/__init__.py
```diff
@@ -19,5 +19,7 @@
 BLOCKED_URLS = [
     'api_token', 'api_regenerate_key',
     'submitservice_view_user_token', 'submitservice_clear_user_token',
+    'edit_profile', 'delete_profile',
+    'auth_password_change', 'aut_password_done',
```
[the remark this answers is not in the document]
A-dead-pixel: The typo still exists.

oioioi/su/middleware.py
A-dead-pixel, line 1: Imho a sanity check whether the effective user isn't a superuser wouldn't hurt.

It would *probably* never happen, but without this person A could know that person B is going to be granted superuser privileges and su to them beforehand.

MasloMaslane: Great catch, added

oioioi/su/middleware.py
```diff
@@ -39,6 +51,44 @@ def _process_request(self, request):
                 request.session[SU_UID_SESSION_KEY],
                 request.session[SU_BACKEND_SESSION_KEY],
             )
+            # Check if the user is contest admin.
+            if not request.session.get(SU_REAL_USER_IS_SUPERUSER, True):
+                # Might happen when switching to a user which then becomes a superuser.
+                if request.user.is_superuser:
+                    return redirect('su_reset')
```
A-dead-pixel, line 58: A check for whether the url is already `su_reset` is needed, otherwise this causes a redirect loop. Also it seems that `su_reset` requires POST, so probably `reset_to_real_user` should be called and a redirect returned (to the main page view ig, as a superuser might have su'ed from outside of a contest, so the contest dashboard won't be proper)..

MasloMaslane: fixed and added a test for this

A-dead-pixel (approved):

PR #405, opened 2024-09-12
Display score change in submissions admin

### LINKED ISSUE #350: [Submission View] Display how many points a contestant gained by submitting this solution
In the current view we only see the raw number of points. It would be way nicer to also display how this submission affected the scoreboard.

Example:

- A submitted solution for problem X, got 0 points (gained points: 0)
- A resubmitted X, got 20 points (gained points: 20)
- A resubmitted X, got 20 points (gained points: 0)
- A resubmitted X, got 100 points (gained points: 80)

MasloMaslane:
Closes #350

### DIFF
--- a/oioioi/contests/admin.py
+++ b/oioioi/contests/admin.py
@@ -852,16 +852,28 @@ class SubmissionAdmin(admin.ModelAdmin):
 
     def score_diff_display(self, instance):
         contest_controller = instance.problem_instance.contest.controller
+        pi_controller = instance.problem_instance.controller
         if not contest_controller.display_score_change() or instance.kind != 'NORMAL':
             return format_html('<span class="text-secondary">-</span>')
 
-        previous_submission = Submission.objects.filter(
-            user=instance.user,
-            problem_instance=instance.problem_instance,
-            kind='NORMAL',
-            date__lt=instance.date,
-        ).order_by('-date').first()
-        return contest_controller.render_score_change(previous_submission, instance)
+        try:
+            previous_submission = pi_controller.get_last_scored_submission(
+                instance.user,
+                instance.problem_instance,
+                before=instance.date,
+            )
+        except Submission.DoesNotExist:
+            previous_submission = None
+        try:
+            curr_submission = pi_controller.get_last_scored_submission(
+                instance.user,
+                instance.problem_instance,
+                before=instance.date,
+                include_current=True,
+            )
+        except Submission.DoesNotExist:
+            curr_submission = None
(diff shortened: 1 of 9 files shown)


metenn (commented):
If I understand correctly, actions like changing the submission kind will also affect the score diff, which might yield unintended results?

I think you accidentally also formatted some existing test code, which, while objectively better, could be confusing for future `git blame`'s and makes this PR's diff a bit harder to navigate

oioioi/contests/admin.py
```diff
@@ -848,6 +850,34 @@ def score_display(self, instance):
...
+            previous_submission = None
+        try:
+            curr_submission = pi_controller.get_last_scored_submission(
+                instance.user,
+                instance.problem_instance,
+                before=instance.date,
+                include_current=True,
+            )
+        except Submission.DoesNotExist:
+            curr_submission = None
```
metenn, line 875: doesn't this always equal `instance`?

oioioi/mp/controllers.py
```diff
@@ -148,17 +149,37 @@ def update_user_result_for_problem(self, result):
             if not best_submission or (score and best_submission_score < score):
                 best_submission = submission
                 best_submission_score = score
+        return best_submission
```
metenn, line 152: consider returning a tuple of `(best_submission, best_submission_score)` to avoid a duplicated redundant DB call later in `update_user_result_for_problem`

metenn (changes requested):

oioioi/contests/admin.py
MasloMaslane: No, for example when the contests takes the highest scored submission as the score for a problem

oioioi/mp/controllers.py
metenn: if we want the function signature of `get_last_scored_submission` to be common among other implementations of it, perhaps dynamically setting an "evaulated score" on the object could do the trick too
e.g. `setattr(best_submission, 'best_submission_score', best_submission_score)`
or, pass the found `ssm` within update_user_result_for_problem to get_last_scored_submission via dynamic kwargs?
If this has a negligible performance difference, keep as is.

oioioi/contests/admin.py
metenn: Right. But then in my opinion `_calculate_score_change` should adhere to contest rules, so an override in MPContestController for instance should return:
`return cls(max(after.value, before.value))`

twalen:

PR #411, opened 2024-10-21
Dismissible GlobalMessage

metenn:
GlobalMessage should be dismissible. This aims to fix that.
- It won't sync between two devices for a client, the alternative requires tracking of every dismissed message for every single account, which requires more server work. For something like this - a message you see once in a blue moon - I think that's a bit overkill.
- The script is inlined in global-message.html to avoid layout shifts (or commonly "flashes") from the page loading well before the script.
- in global-message.html, the context variable `global_message` is also now passed through `|safe` to allow for more customizable messages via HTML. This isn't necessarily an XSS attack vector, as GlobalMessage is only changeable if `user.is_superuser = True`, and superusers already have full access to many other system components.


twalen:

PR #284, opened 2023-11-01
Extra text on contest pages

MasloMaslane:
Closes #281 and #279.
Test run description was written by OI's Jury.

Public messages can be added in:
- `forum`: Forum view, thread view
- `contests`: Files view, My submissions view, Submit view
- `rankings`: Ranking view
- `questions`: Question list view, New question view

All of these messages can be edited in web UI or in contest controller

PR #412, opened 2024-10-23
Public message in Submission view

MasloMaslane:
As per OI's Jury request, I added a public message in Submission view (continuation of https://github.com/sio2project/oioioi/pull/284)


twalen (approved):
Looks OK, but this is getting out of control
image

PR #423, opened 2024-11-19
[Forum] Display who upvoted/downvoted each post. Closes #346

### LINKED ISSUE #346: [Forum] Display who upvoted/downvoted each post
It might be just a list displayed when hovered over the plus/minus.

stopnoanime:
Displays usernames of users who upvoted/downvoted each post in a tooltip.
Only 10 usernames for each post are queried and displayed to improve performance.
I'm using prefetch_related to query necessary reactions for all posts with a single query.

### To Do:
- Add tests
- Improve translation ( I don't know if I should also change files in /locale )


oioioi/forum/templatetags/display_reacted_by.py
```diff
@@ -0,0 +1,23 @@
+from django import template
+from django.utils.translation import gettext as _
+from oioioi.base.utils import get_user_display_name
+from django.conf import settings
+
+register = template.Library()
+
+@register.simple_tag
+def display_reacted_by(post, rtype):
+  # post needs to be queried with prefetch_reacted_by
```
twalen, line 10: I suggest to add sanity check if rtype is "upvote" or "downvote" and tigger an error if incorrect rtype is used.

oioioi/forum/templatetags/display_reacted_by.py
```diff
@@ -0,0 +1,23 @@
...
+
+  output = ', '.join([
+    get_user_display_name(reaction.author)
+    for reaction in getattr(post, rtype.lower() + 'd_by')
```
twalen, line 14: why `.lower()`, since it is an internal tag we can expect that someone will use it with proper rtype

stopnoanime:
This PR has a bug that causes the latest posts view to not load. I turned the PR to a draft while I investigate why the tests didn't detect this.

oioioi/forum/templatetags/display_reacted_by.py
stopnoanime: Added the check.

stopnoanime:
Added a test that checks the latest post view with reactions enabled.
Added two dictionaries that map reactions to necessary attribute names so the code doesn't rely on string manipulations anymore. Also moved prefetching reacted_by into the Post model so there is no need to call a separate function to do so.

oioioi/forum/models.py
```diff
@@ -192,22 +193,26 @@ class Post(models.Model):
...
                 }
                 qs = qs.annotate(**reaction_count_agg)
 
+            max_count = getattr(settings, 'FORUM_REACTIONS_TO_DISPLAY', 10)
+            for rtype, attr_name in POST_REACTION_TO_PREFETCH_ATTR.items():
+                qs = qs.prefetch_related(
+                    models.Prefetch(
+                        'reactions', 
+                        to_attr=attr_name,
+                        queryset=PostReaction.objects.filter(type_of_reaction=rtype).select_related('author')[:max_count],
```
twalen, line 212: it would be nice to add some ordering here, otherwise we can get random results (especially important during tests): i.e. `order_by('-pk')`

also you can add Meta class to PostReaction like:
```
class Meta:
        ordering = ['-pk']
```

twalen (changes requested):
Looks ok, but please add ordering + modify tests

oioioi/forum/models.py
stopnoanime: Done.

stopnoanime:
Tests started failing due to changing Meta options on PostReaction, will need to figure it out.

stopnoanime:
Adding the migration fixed the issue. I didn't realize a change like this would require a migration.

stopnoanime:
Decided to undo the meta ordering change and migration. The meta ordering seems unnecessary and adding it requires a migration file.

twalen (approved):

twalen (approved):

PR #431, opened 2024-11-28
Aggregated tag proposals

### LINKED ISSUE #419: Feature Request: Aggregating Algorithm Tag Proposals.
As of right now, each Tag Proposal shows up separately in the admin panel of given task. 

Proposed aggregation is, if tag `BFS` has been proposed 12 times, instead of displaying `BFS` 12 times to display `BFS` with the number 12 next to it. Then clicking on the tag could show the usernames of all users who proposed that tag (which may be useful information to deal with saboteurs).

I believe there also should be a way regular users to see the aggregated tag proposals, as well as an option to display them under 'Tags' after official ones when viewing the task database as a regular user.

This feature allows for a sensible implementation of searching the task database for tasks based on both official and proposed tags, rather than just the official ones.

segir187:
(Partly) Resolves #419.
Draft pull request for implementation of AggregatedAlgorithmTagProposal and AggregatedDifficultyTagProposal classes.

The classes as of right now do nothing other than being updated each time their respective TagProposal has a proposal added or removed. The end-goal of implementing these classes is to enable filtering in the by proposed AlgorithmTags in the Problemset, the mid-goal being showing AggregatedTagProposals in the admin menu in the settings for a specific task.

The original idea was to only create AggregatedAlgorithmTagProposal, without AggregatedDifficultyTagProposal, however it seemed improper to only aggregate one of the two, so I did both, especially since the code is quite directly analogous.

An additional idea I got for how AggregatedDifficultyTags could be used is to display the median DifficultyTag next to in the Problemset if there is no officially assigned one. However, that would require adding numeric values to DifficultyTags.


segir187:
Quite importantly, within this PR, there's a data migration that creates AggregatedTagProposal tables from preexisting TagProposal tables.

oioioi/problems/models.py
```diff
@@ -885,6 +886,62 @@ class Meta(object):
...
+        aggregated_difficulty_tag_proposal = AggregatedDifficultyTagProposal.objects.get(problem=problem, tag=tag)
+        aggregated_difficulty_tag_proposal.amount -= 1
+        
+        if aggregated_difficulty_tag_proposal.amount == 0:
+            aggregated_difficulty_tag_proposal.delete()
+        else:
+            try:
+                aggregated_difficulty_tag_proposal.save()
+            except ValidationError as e:
+                raise RuntimeError(
```
twalen, line 919: I suggest to log an error instead of raising an exception

oioioi/problems/migrations/0033_populate_aggregated_tag_proposals.py
```diff
@@ -0,0 +1,44 @@
...
+    AggregatedAlgorithmTagProposal = apps.get_model('problems', 'AggregatedAlgorithmTagProposal')
+
+    AggregatedDifficultyTagProposal.objects.all().delete()
+    AggregatedAlgorithmTagProposal.objects.all().delete()
+
+    difficulty_data = (
+        DifficultyTagProposal.objects.values('problem', 'tag')
+        .annotate(amount=models.Count('id'))
+    )
+    for entry in difficulty_data:
```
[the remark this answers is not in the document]
twalen: https://docs.djangoproject.com/en/5.1/ref/models/querysets/#bulk-create
https://docs.djangoproject.com/en/5.1/howto/writing-migrations/#migrating-data-between-third-party-apps

oioioi/problems/models.py
```diff
@@ -885,6 +886,62 @@ class Meta(object):
...
 
 
+
+@receiver(post_save, sender=DifficultyTagProposal)
+def increase_aggregated_difficulty_tag_proposal(sender, instance, created, **kwargs):
+    problem = instance.problem
+    tag = instance.tag
+    
+    if created:
+        aggregated_difficulty_tag_proposal = AggregatedDifficultyTagProposal.objects.get_or_create(
```
twalen, line 896: do we need locks?

oioioi/problems/tests/test_tag_proposals.py
```diff
@@ -0,0 +1,337 @@
...
+                'user': 'test_admin',
+                'problem': '0',
+            },
+        )
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(AlgorithmTagProposal.objects.count(), 2)
+        self.assertEqual(AggregatedAlgorithmTagProposal.objects.count(), 2)
+        self.assertEqual(DifficultyTagProposal.objects.count(), 1)
+        self.assertEqual(AggregatedDifficultyTagProposal.objects.count(), 1)
+        self.assertTrue(
```
twalen, line 167: looks very complicated, maybe add some utility function that reads tag proposals and returns a dictionary with {name: amount}

twalen (changes requested):
Looks good, but consider my comments.

segir187:
A thing which may be important to address is that I added a test for created data migration, however it uses regular pytest-django structure, since after reviewing multiple possible solutions through links given to me by @twalen, I concluded that none of them were at all needed to test this data migration. The most sensible option was using django-test-migrations, however it neccesitates installing an additional app to the whole project and it seemingly (?) does not play nice with pytest-django. I opted to simply test if the function doing the data migration, defined in the 0033 migration file, migrates data correctly by invoking it directly in the code, as it is essentially synonymous to invoking the whole migration, while not needing any extra apps to be installed and issues in compatibility to be worked through.

segir187:
As for the unit test failure, I've no idea what causes it, as in, I switched to commit [b385cd7](https://github.com/sio2project/oioioi/pull/431/commits/b385cd7bc2e03ad5df2d6c1263f2b90da5c3dbd1), before any changes were made and, at least locally, on my computer, I had the exact same test failure.

I was also unable to find any sort of interaction between my changes and the failed tests.

segir187:
Dopiero teraz zauważyłem, że lokalnie nie przechodziły mi 4 testy, a tutaj nie przeszedł tylko 1, który chyba jest jednym z owych nieprzechodzących testów.

segir187:
Just now noticed, that locally, for consecutive tests failed for me and after the push only one of them fails. Moreover, commit 8ee264f also passed with no errors.

oioioi/problems/models.py
segir187: Completely rewrote these functions, any caught exceptions in decreasing aggregated tags are logged.

twalen (approved):
Looks good

MasloMaslane (dismissed):

segir187:
Accidentally pushed something onto this branch. Unsure how to undo this.

segir187:
Seemingly successful rollback to the reviewed version!

twalen (approved):
OK

PR #433, opened 2024-11-30
Fix and improve the docker setup

otargowski:
This PR fixes building the docker image, as after the recent change in filetracker packaging one must explicitly specify `[server]` to install dependencies for running an ft server. Without this, the filetracker server can't start when it is supposed to receive sandboxes.

I opted for putting this only in the Dockerfile, as manual setups may not need to run an ft server.

Additionally, I made the filetracker data live in a docker volume when running via `easy_toolbox`/`docker-compose`.
Without this, after one runs `./easy_toolbox.py down`, they get and inconsistent state: the db remains, but files on filetracker are gone.
I added the `wipe` command to `easy_toolbox` to allow for clearing the volumes and marked the `down` one as non-dangerous.

I also removed the `version` entries from docker-compose files due to this warning:
```
WARN[0000] .../oioioi/docker-compose-dev.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion
```
The issue mentioned in a related comment was resolved a few years ago.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,7 +60,7 @@ RUN pip3 install --user bsddb3==6.2.7
 WORKDIR /sio2/oioioi
 
 COPY --chown=oioioi:oioioi setup.py requirements.txt ./
-RUN pip3 install -r requirements.txt --user
+RUN pip3 install -r requirements.txt --user filetracker[server]
 COPY --chown=oioioi:oioioi requirements_static.txt ./
 RUN pip3 install -r requirements_static.txt --user
 
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -1,5 +1,3 @@
-version: '2.4'
-# https://github.com/docker/cli/issues/1293
 services:
   db:
     image: library/postgres:12.2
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -60,7 +60,7 @@ RUN pip3 install --user bsddb3==6.2.7
 WORKDIR /sio2/oioioi
 
 COPY --chown=oioioi:oioioi setup.py requirements.txt ./
-RUN pip3 install -r requirements.txt --user
+RUN pip3 install -r requirements.txt --user filetracker[server]
```
Iteron-dev, line 63: I'm not sure if this is the best approach:
- the version range for `filetracker` is not specified.
- In `oioioi`, there is a `setup.py` file where this requirement is defined. 

I have already addressed this issue in the #434 , due to the Python update.

Iteron-dev, line 63: It would be better to specify the version range for `filetracker`.

otargowski: But doesn't pip satisfy tbose constraints already?
In requirements.txt there is `-e .`, which afaik include setup.py constraints, which contain the version range.
And it's better to not have to update in two places.

Iteron-dev: Yeah, you’re right.

Iteron-dev (approved):

twalen (approved):
Looks good

PR #434, opened 2024-12-01
Update Python version to 3.10 in Dockerfile & Github actions

### LINKED ISSUE #418: Update Python version in Dockerfile & Github actions
Currently Python 3.8 is used for both Docker & Github actions (like pytest, nightbuild, cypress).

Python 3.8 is unsupported right now (end of life on 07 Oct 2024 - https://endoflife.date/python)

Consider updating to 3.9 or newer.
Please make sure that new version is compatible with current python dependencies (or update them)

Iteron-dev:
Resolve #418.


MasloMaslane:
@twalen could we wait a bit before merging this?

setup.py
```diff
@@ -64,7 +64,7 @@
     # Used in quizzes module for adding new quizzes.
     "django-nested-admin>=4.0,<4.1",
     # SIO2 dependencies:
-    "filetracker>=2.1,<3.0",
+    "filetracker[server]>=2.2.0,<3.0",
```
MasloMaslane, line 67: Is it really required to add `filetracker[server]` to setup.py requirements? Production code on Szkopul requires only client code, as filetracker server is running on another system. And installing with server dependencies in prod crashed because of some version conflicts. Couldn't it be handled like in Olaf's PR https://github.com/sio2project/oioioi/pull/433 ?

Dockerfile
```diff
@@ -54,7 +54,7 @@ USER oioioi
 ENV PATH $PATH:/home/oioioi/.local/bin/
 
 ENV BERKELEYDB_DIR /usr
-RUN pip3 install --user psycopg2-binary==2.8.6 twisted uwsgi
+RUN pip3 install --user psycopg2-binary==2.9.10 twisted uwsgi
```
MasloMaslane, line 57: Could you explain a bit why this version bump is required?

MasloMaslane (changes requested):

setup.py
Iteron-dev: Good point! I'll remove the `[server]` optional dependencies from `setup.py` and add them to the `Dockerfile`.

Dockerfile
Iteron-dev: Actually, this version bump isn't required. Initially, there was a plan to upgrade to Python 3.11, but on Python 3.10, psycopg 2.8.6 works well.

twalen (approved):
Looks good, as for me it is ready to be merged.

MasloMaslane (approved):

Issue #427, opened 2024-11-25
Unable to open tasks in a new tab


wraczuk:
When searching for tasks, trying to open a task in a new tab (e.g., with middle-click or "Open in new tab") duplicates the current page instead of opening the selected task.
![image](https://github.com/user-attachments/assets/0be694b7-abcb-42cd-99f3-ed18fa63447e)

PR #437, opened 2024-12-02
Add opening tasks in new tab from search bar (#427)

wraczuk:


oioioi/base/static/js/bootstrap-typeahead.js
twalen, line 1: have you updated this file to the latest version from https://github.com/bassjobsen/Bootstrap-3-Typeahead/blob/master/bootstrap3-typeahead.js ?
it seems that there multiple versions with "v4.0.2"

twalen (approved):
Looks ok, just please confirm my question about version of typahead.js library

oioioi/base/static/js/bootstrap-typeahead.js
wraczuk: Yes, we have updated to the latest version (which you linked).

PR #438, opened 2024-12-07
Add ordering for DifficultyTag and DiffultyTagLocalization

### LINKED ISSUE #344: DifficultyTag's aren't sorted
Seems like somebody forgot `ordering = ['pk']` in `Meta`
https://github.com/sio2project/oioioi/blob/5414a6ce7dbe790905114895e8b2415674a70f43/oioioi/problems/models.py#L818-L838
same goes for `DifficultyTagLocalization`, which is why this code returns unsorted rows (per SQL specs rows are not automatically sorted by any field)
https://github.com/sio2project/oioioi/blob/5414a6ce7dbe790905114895e8b2415674a70f43/oioioi/problems/views.py#L556
so another fix would probably be adding an `.order_by()` here?

perhaps the ordering should be more explicit in the first place rather than relying on the autogenerated `pk` field?

reivvax:
Closes #344


twalen (approved):
Looks good

PR #447, opened 2024-12-16
[Demo] Add a simple oauth implementation via django-oauth-toolkit

gkju:
Adding barebones oauth support is pretty trivial when using django-oauth-toolkit and this commit demonstrates the MVP. In a debug environment as per the [docs](https://django-oauth-toolkit.readthedocs.io/en/latest/rest-framework/getting_started.html#step-2-create-a-simple-api) one can use `/o/applications/` to register an oauth2 app with ex;
```
Client Type: confidential
Authorization Grant Type: Resource owner password-based
```
then retrieve an access token for admin (after replacing `<client_id>`, `<client_secret>` with the appropriate values) using `curl`;
```sh
curl -X POST -d "grant_type=password&username=admin&password=admin" -u"<client_id>:<client_secret>" http://localhost:8000/o/token/
```
and authenticate API calls via the access token (using the bearer scheme) which can be refreshed with the refresh token as in the spec. Other oauth2 flows work out of the box as well. In a production environment, I would implement special admin pages for managing the applications, and add config options for enabling oauth/defining clients. The toolkit package also adds some new migrations for the oauth-related models which have to be applied in current installations.

## The things to consider are;
- Using OIDC
- What should be included in a default installation
- What means of configuration would be preferred (defining clients statically in config files/managing them via a webui)

This pull request would be very useful for the upcoming TAG semester, in my opinion, as it greatly simplifies creating discrete applications that interconnect with oioioi installations.

### DIFF
--- a/oioioi/default_settings.py
+++ b/oioioi/default_settings.py
@@ -286,6 +286,7 @@ INSTALLED_APPS = (
     'oioioi.quizzes',
     'oioioi._locale',
     'oioioi.interactive',
+    'oioioi.oauth',
 
     'djsupervisor',
     'registration',
@@ -313,6 +314,7 @@ INSTALLED_APPS = (
 
     'nested_admin',
     'coreapi',
+    'oauth2_provider',
     'rest_framework',
     'rest_framework.authtoken',
 
@@ -845,6 +847,7 @@ REST_FRAMEWORK = {
     'DEFAULT_AUTHENTICATION_CLASSES': (
         'rest_framework.authentication.TokenAuthentication',
         'rest_framework.authentication.SessionAuthentication',
+        'oauth2_provider.contrib.rest_framework.OAuth2Authentication',
     ),
     'DEFAULT_SCHEMA_CLASS': 'rest_framework.schemas.coreapi.AutoSchema'
 }
--- a/oioioi/oauth/urls.py
+++ b/oioioi/oauth/urls.py
@@ -1,0 +1,33 @@
+import oauth2_provider.views as oauth2_views
+from django.conf import settings
+from django.contrib import admin as django_admin
+from django.views import i18n
+from django.urls import include, re_path
+
+app_name = 'oauth'
+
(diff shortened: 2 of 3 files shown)


oioioi/deployment/settings.py.template
```diff
@@ -598,3 +598,23 @@ CONTEST_ADMINS_CAN_SU = False
...
 ALLOW_ONLY_GET_FOR_SU_CONTEST_ADMINS = True
+
+# OAuth2 support
+#
+# INSTALLED_APPS = (
+#    'oauth2_provider',
+#    'oioioi.oauth',
+# ) + INSTALLED_APPS
+#
+# REST_FRAMEWORK = {
```
twalen, line 609: this key is already defined in default_settings.py, maybe we can just add `'oauth2_provider.contrib.rest_framework.OAuth2Authentication' key

oioioi/oauth/tests/tests.py
```diff
@@ -0,0 +1,94 @@
...
+from django.urls import NoReverseMatch, reverse
+from django.contrib.auth import get_user_model
+from oauth2_provider.models import Application, AccessToken
+from django.utils import timezone
+from rest_framework.test import APIClient
+import json
+
+from oioioi.base.tests import TestCase
+
+User = get_user_model()
```
twalen, line 25: this overrides import from the line 7

twalen (changes requested):
Looks good, I've added minor comments.

oioioi/oauth/tests/tests.py
gkju: Removed multiple unnecessary imports

oioioi/test_settings.py
```diff
@@ -125,3 +127,16 @@
     'handlers': ['console'],
     'level': 'INFO',
 }
+
+REST_FRAMEWORK = {
```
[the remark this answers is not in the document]
gkju: Done (see previous)

PR #448, opened 2024-12-17
Fix `get_queryset()` on Problem admin

MasloMaslane:
Super user should have access to all problems and `problems.problems_db_admin` should have access to any public problems, not only problems without a contest.


twalen (approved):
OK

PR #449, opened 2024-12-22
Ordering of DifficultyTags

### LINKED ISSUE #344: DifficultyTag's aren't sorted

reivvax:
closes #344


MasloMaslane (changes requested):
Could you remove commit https://github.com/sio2project/oioioi/pull/449/commits/27a3d990b38ce40bfe9835f64e38cc7b1f44c8b7? It seems like you merged squashed master, it makes this PR unreadable

twalen (approved):

twalen (changes requested):
PR looks good, but please change migration number to 0036 (0035 is already used in the master).

PR #450, opened 2024-12-22
Fix flaky test TestProblemsetFilters::test_filters. Closes #440

### LINKED ISSUE #440: Flaky tests
Sometimes (in 5-10% of GitHub Actions runs), unit and nightly tests fail. I found two flaky tests:

- Nightly tests: [link](https://github.com/sio2project/oioioi/actions/runs/12133571941) - FAILED oioioi/problems/tests/test_problemset.py::TestProblemsetFilters::test_filters
- Unit tests: [link](https://github.com/sio2project/oioioi/actions/runs/12123980565) - FAILED oioioi/sinolpack/tests.py::TestSinolPackage::test_single_file_replacement

stopnoanime:
### The problem
This test used to check for the presence of a problem by asserting that the response HTML contained the problem name.
However, the response also included an HTML form with a generated CSRF token, sometimes this token would contain the problem name (test problems have short names like `aaa` so it wasn't that rare).
This caused the test to randomly fail.

### Solution
Now the test expliclitly searches for a HTML tag with the necessary problem name.

### DIFF
--- a/oioioi/problems/tests/test_problemset.py
+++ b/oioioi/problems/tests/test_problemset.py
@@ -219,7 +219,9 @@ class TestProblemsetFilters(TestCase):
             self.assertEqual(response.status_code, 200)
 
             for problem in self.problems:
+                problemTag = f"<td>{problem}</td>"
+
                 if problem in filtered:
-                    self.assertContains(response, problem)
+                    self.assertContains(response, problemTag, html=True)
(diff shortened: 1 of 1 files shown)


twalen (approved):
Looks Good. But I wonder how many the same kind of errors still remain in the code.

PR #461, opened 2025-01-20
Mock csrf token in all tests. Fixes #454

### LINKED ISSUE #454: Find and fix more flaky tests.
Tests that use assertNotContains can randomly fail (see #450) due to the values that appear in the csrf token.

I want to find all tests that use assertNotContains and can fail and fix them.

stopnoanime:
Instead of changing all test that use assertNotContains, we can just mock the csrf token for testing so it always has the same value. Thanks to this the generated HTML does not contain randomly generated strings, so tests should not fail randomly anymore.

### DIFF
--- a/oioioi/base/tests/__init__.py
+++ b/oioioi/base/tests/__init__.py
@@ -4,6 +4,7 @@ from contextlib import contextmanager
 
 import pytest
 import urllib.parse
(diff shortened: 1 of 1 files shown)


twalen (approved):
Looks good

PR #467, opened 2025-02-25
Action speedup

stopnoanime:
Managed to speedup test actions by running django tests in parallel. Also made the cypress task actually work again.

Trying to cache pip packages didn't improve performance (when cache was used it saved ~5s of runtime, but saving the cache took ~2 minutes every time it needed to be regenerated). Trying to use containers with preinstalled apt dependencies improved performance only slightly, but introduced multiple other issues.

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -28,9 +28,8 @@ jobs:
           python easy_toolbox.py up --no-input
 
       - name: Wait for migrations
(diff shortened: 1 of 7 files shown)


stopnoanime:
Also removed the free pascal compiler because afaik it is not needed anymore.

twalen (approved):
Looks very good.
Thank you for all investigations.

Issue #471, opened 2025-02-26
Add from problemset does not copy timelimits


twalen:
Scenario:

add problem using
- Tasks -> Add problem -> Add from problemset -> I have a secret key

current behaviour:
- task is copied into a contest, but the timelimits are set to default (every test has timelimit = 10s)

expected behaviour:
- timelimits are the same as in the original task

kamilszymczak1:
Copying problems from URL keys does not copy contest-specific time or memory limits. Instead, the limits are set to the default package settings. In order to copy time limits with contest-specific settings, one has to go to a contest's problem list and choose the 'Attach to another contest' option.

I've added a note about this displayed to users when adding a problem from a URL key.

PR #480, opened 2025-03-06
Add a note about copying limits

kamilszymczak1:
Copying problems from url keys does not copy contest-specific time or memory limits. Instead, the limits are set to the default package settings. In order to copy time limits with contest-specific settings, one has to go to a contest's problem list and choose the 'Attach to another contest' option.

This pull request adds a note about it displayed when adding problems through url keys.

It is a solution to the issue[ #471](https://github.com/sio2project/oioioi/issues/471).

### DIFF
--- a/oioioi/problems/forms.py
+++ b/oioioi/problems/forms.py
@@ -105,7 +105,7 @@ class ProblemSiteForm(forms.ModelForm):
 
 class ProblemsetSourceForm(forms.Form):
     url_key = forms.CharField(label=_("Enter problem's secret key"),
-                              help_text=("Please note that the time and memory limits will be copied\
+                              help_text=_("Please note that the time and memory limits will be copied\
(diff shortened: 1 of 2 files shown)


oioioi/problems/forms.py
```diff
@@ -104,7 +104,14 @@ class Meta(object):
 
 
 class ProblemsetSourceForm(forms.Form):
-    url_key = forms.CharField(label=_("Enter problem's secret key"), required=True)
+    url_key = forms.CharField(label=_("Enter problem's secret key"),
+                              help_text=("Please note that the time and memory limits will be copied\
```
twalen, line 108: you need to use `_` function to be able to translate this text, i.e.: `_("Please note ...`

kamilszymczak1: Done :+1:

twalen (approved):

Issue #453, opened 2025-01-09
Django 5 migration


Iteron-dev:
After updating Python to 3.10, it becomes possible to upgrade Django from 4.2 to 5.x. Considering that Django 4.2 will reach the end of security support on April 30, 2026, upgrading to Django 5 sooner is a good idea.

Since Django 5.2 is expected to be released in April 2025, ensure that the migration to Django 5.1 addresses all features deprecated in Django 5.2. For details, see: https://docs.djangoproject.com/en/dev/releases/5.2/#deprecated-features-5-2.

PR #483, opened 2025-03-12
Django 5 migration

Iteron-dev:
Resolve https://github.com/sio2project/oioioi/issues/453.


oioioi/workers/urls.py
```diff
@@ -1,10 +1,11 @@
...
 from django.urls import re_path
 
 from oioioi.workers import views
 
 app_name = 'workers'
 
 urlpatterns = [
-    re_path(r'^workers/$', views.show_info_about_workers, name='show_workers'),
+    path('workers/', views.show_info_about_workers, name='show_workers'),
     re_path(r'^workers/load.json$', views.get_load_json, name='get_load_json'),
```
MasloMaslane, line 10: Is there a reason this is not changed?

MasloMaslane (commented):
I have a few questions (in context of upgrading sio2.mimuw to Django 5):
- why are some migrations changed? Is it possible to do this automatically?
- should in every admin `.short_description` be replaced with `@admin.display(description=...)`?

oioioi/contests/migrations/0021_rename_contests_contestview_user_id_timestamp_a0151775_idx_contests_co_user_id_9b1d73_idx_and_more.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 4.2.20 on 2025-04-15 22:08
```
twalen, line 1: @MasloMaslane This is the biggest concern in the migration to Django 5.2.
Due to change of index handling we need to fix the names of the indexes
(see https://adamj.eu/tech/2020/07/27/how-to-modernize-your-django-index-definitions/)

Default migration method requires long downtime due to rebuilding of the indexes, this suppose to be zero-downtime method, but we need to make sure tha the "old_names" are correct.

oioioi/workers/urls.py
Iteron-dev: This change was automatically made by [django-upgrade](https://github.com/adamchainz). I could change the remaining re_path to path manually if needed.

twalen (approved):
Looks good.

Issue #459, opened 2025-01-20
Optional way to see aggregated tag proposals in probset for all users.


segir187:
Before implementing filtering by tag proposals, a way to see them in the probset should be implemented.

I think filtering by tags should only include aggregated tag proposals when seeing them is turned on.

PR #489, opened 2025-03-18
Searchbox bugfix

### LINKED ISSUE #477: Bugged out filtering by difficulty tags
I encountered this problem while testing out a feature I was coding.

After choosing to filter by the "easy" difficulty tag I ended up on url `http://localhost:8000/problemset/?q=&difficulty=Easy` with the page looking as I expected, like this:

Image

Then, after clicking on the searchbox again and trying to filter again without adding tags I ended up on url `http://localhost:8000/problemset/?q=&difficulty=easy` with the page looking like the following:

Image

When adding an additional tag to filter by, the `Easy` in the url also changes to `easy` and the box signifying that it is being included in the filters disappeared.

I plan on fixing this myself as soon as I finish the feature I am currently working on, as I need to understand the inner workings of the searching algorithm to implement my next planned feature.

### LINKED ISSUE #485: Filter by infinite copies of same Difficulty Tag
It is possible to filter by the same tag an unlimited amount of times.

How to reproduce the bug:
1. Have a Difficulty tag with a unique name.
2. Put enough of its name into the searchbox, so that it is the only tag that shows up.
3. Press the down arrow key multiple times - each press of the button adds another copy of the tag icon beneath the search box and to the url after clicking the search button.

Picture of how it looks before clicking the search button:

Image

And after:

Image

### LINKED ISSUE #486: Filtering by tags in probset with show_tags = false.
Fitering by tags in probset is possible and works even if displaying them is disabled.

Image

segir187:
Fixes #477. Fixes #485. Fixes #486.

Fix of three bugs related to the problemset searchbox I found while implementing #459 .


oioioi/problems/static/common/tag_selection.js
```diff
@@ -116,10 +116,10 @@ function init_search_selection(id) {
                 } else if (item.trigger !== 'problem') {
                     // At this point for anything other than a problem we
                     // want to create a search tag
-                    const value = item.value || item.name;
+                    const value = item.value; // IS THIS CHANGE GOOD?
```
segir187, line 119: Previously this value defaulted to `item.name` if `item.value` was not given.

As I am pretty sure that all places in the code where `item.value` was previously not given I changed it to be given, is the removal of this value defaulting to `item.name` a good change or is it still unnecessary?

oioioi/problems/views.py
```diff
@@ -1181,9 +1181,11 @@ def get_search_hints_view(request, view_type):
 
     result = []
     result.extend(list(get_problem_hints(query, view_type, request.user)))
-    result.extend(get_algorithm_and_difficulty_tag_hints(query))
-    result.extend(get_nonselected_origintag_hints(query))
-    result.extend(get_origininfovalue_hints(query))
+
+    if settings.PROBLEM_TAGS_VISIBLE or request.user.is_superuser:
```
segir187, line 1185: This fix only ensures that hints related to tags are not given to non-superusers. It is still possible for any user to manually filter by the tags they want by changing the URL.

Should I make it so that it is impossible for non-superusers to filter by tags in such a way if they are hidden?

Should I make it so that superusers are also unable to filter by hidden tags, as the tags still don't show up next to tasks for them when `PROBLEM_TAGS_VISIBLE = False`?

oioioi/problems/tests/test_tags.py
```diff
@@ -159,7 +159,39 @@ class TestProblemSearchHintsTags(TestCase, AssertContainsOnlyMixin):
     def get_query_url(self, parameters):
         return self.url + '?' + urllib.parse.urlencode(parameters)
 
-    @override_settings(LANGUAGE_CODE="en")
+    @override_settings(LANGUAGE_CODE="en", PROBLEM_TAGS_VISIBLE=False)
+    def test_search_no_hints_tags(self):
```
segir187, line 163: I added this test to make sure that with `PROBLEM_TAGS_VISIBLE=False` the hints do not show up. Should I duplicate each test that fails with `PROBLEM_TAGS_VISIBLE=False` like this?

twalen (approved):

PR #499, opened 2025-04-02
Show problem limits

### LINKED ISSUE #167: Show time/memory limits
Add an option to show time/memory limits in the problem list. This should be a switch in the contest settings. In case of different limits in different tests, it should show "A-B seconds" where A is the minimum and B is the maximum over all the TLs.

reivvax:
closes #167


oioioi/contests/migrations/0020_limitsvisibilityconfig.py
twalen, line 1: please change migration number to 0021 and update depdendencies

twalen (approved):

MasloMaslane (changes requested):
Why only c++ and python?

reivvax:
> Why only c++ and python?

@MasloMaslane it was decided together with prof. @twalen. Two reasons:
- there are two most used languages
- we did not want to clutter UI too much

twalen:
> > Why only c++ and python?
> 
> @MasloMaslane it was decided together with prof. @twalen. Two reasons:
> 
> * there are two most used languages
> * we did not want to clutter UI too much

I guess that we can add C as well (in OI it might be more popular than Python)

MasloMaslane (dismissed):

twalen (approved):
Looks good.

PR #503, opened 2025-04-08
Filter by tag proposals

### LINKED ISSUE #420: Feature Request: Using proposed algorithm tags for filtering tasks.
Feature meant to make it easier for users of platforms like Szkopuł to find tasks with few or none admin-given tags, when searching for tasks using a chosen algorithm.

As of now, as far as I'm aware, the proposed tasks are only visible to admins and the only way for users to interact with them is by proposing tags for a task. This would make better use of them.

Note: implementing this makes sense only after the implementation of issue #419.

### LINKED ISSUE #487: Bugfixes and changes to task filtering in probset
This is an issue to assign as parent to other issues pertaining to the searchbox in the probset for organisation purposes, as there are a lot of them.

segir187:
Closes #420. Closes #487.

Make it possible to search by algorithm tag proposals.

Includes wording change for the checkbox controlling tag proposals so Transifex will need to be updated if the change is kept.


twalen (approved):
looks good

PR #506, opened 2025-04-16
add test option to check if variables in template tags are defined

### LINKED ISSUE #504: Undefined Variables in Django Templates Evaluated as 0 or None Without Raising an Error
In Django templates, when referencing a variable that is not defined in the context, the template engine silently evaluates the undefined variable as None or 0 without raising an error. This behavior can lead to unexpected results.

I want to update tests to raise an error when undefined variables are used in templates.

Atanazyy:
closes #504


MasloMaslane:
Is there a reason why this shouldn't be always checked?

Atanazyy:
Currently some test fail with this flag, because for example in ```py
<a href="{{ item.url }}" id="{{ item.name }}" {{ item.attrs }} class="list-group-item list-group-item-action {% active_url item.url %}">
``` because `item.name` might not be defined and everything works fine. I want to fix things like that and then set this as default check

Atanazyy:
image
For now quite a lot of tests fail

twalen (approved):

PR #513, opened 2025-05-20
Mass create tool

### LINKED ISSUE #512: Tool to create model objects for testing purposes
I needed such a tool to benchmark my #503 PR and since I implemented it, it seems like it may be useful for testing features relating to problem tagging and is easily extendable to create objects of even more models.

segir187:
Resolves #512. I created this tool for benchmarking #503 and generalised it so it can be a stand-alone feature. I was missing a tool like this while implementing my previous issues. There is no commit history as it is present in the #503 PR.

Currently it is able to create Problems with associated ProblemSites (necessary for them to show up in the Probset), as well as Users and Algorithm and Difficulty Tags, TagThrough records and TagProposals.

### DIFF
--- a/oioioi/problems/models.py
+++ b/oioioi/problems/models.py
@@ -169,9 +169,9 @@ class Problem(models.Model):
         )
 
     def __str__(self):
-        return u'%(name)s (%(short_name)s)' % {
-            u'short_name': self.short_name,
-            u'name': self.name,
+        return '%(name)s (%(short_name)s)' % {
+            'short_name': self.short_name,
+            'name': self.name,
         }
 
     def save(self, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


oioioi/problems/models.py
```diff
@@ -855,7 +855,7 @@ class DifficultyTagThrough(models.Model):
 
     # This string will be visible in an admin form.
     def __str__(self):
-        return str(self.tag.name)
+        return str(self.problem.name) + u' -- ' + str(self.tag.name)
```
twalen, line 858: u prefix is no longer needed in Python 3.x (this is the default mode)

twalen (changes requested):
Looks good, but please add some automated tests (i.e. few runs of the script with different arguments to make sure that some new objects are created)

oioioi/problems/models.py
segir187: I changed this for all `__str__` methods in `models.py`.

twalen (approved):
Looks good! Thank you for the changes.

Issue #285, opened 2023-11-07
Add a dark theme


patgra1410:
I propose to add a dark theme to SIO, I don't think there is much more to add. Many people complete OI problems at night and would much rather not get their eyes burnt out.

Antix7:
Rel

MasloMaslane:
Adding dark theme isn’t currently a priority, here are the reasons: #106. Personally, I’m using [this](https://github.com/darkreader/darkreader) extension for dark theme and it works quite well on SIO

Issue #106, opened 2022-11-23
Introduce dark mode


SmolSir:
With the Bootstrap updated, maybe there is now a chance for this?

DietPawel:
@MasloMaslane said:
```quote
I’m using [this](https://github.com/darkreader/darkreader) extension for dark theme and it works quite well on SIO
```
I think that we could inform our users about the possibility of using such an extension. I would say it is a good compromise between helping users and making our UI more complex to maintain. Feel free to recommend other tools, which work well or/and proper notification text at least in English or/and even a pull request!

PR #517, opened 2025-05-21
Add dark mode (#106, #285)

wraczuk:
Adds dark mode utilising Dark Reader library.


twalen (changes requested):
Please resolve merge conflicts & move darkreader into NPM.

oioioi/base/templates/base.html
```diff
@@ -2,9 +2,27 @@
...
+                if (localStorage.getItem("dark-mode") === "enabled") {
+                    DarkReader.disable();
+                    localStorage.setItem("dark-mode", "disabled");
+                } else {
+                    DarkReader.enable();
+                    localStorage.setItem("dark-mode", "enabled");
+                }
+            });
+        });
+    </script>
```
stopnoanime, line 24: I think this should be moved to darkreader.js; there's no need for it to be here. Also, after moving it, assigning DarkReader to the window object will no longer be necessary, which is a plus.

twalen (approved):

PR #519, opened 2025-05-28
Update Python version to 3.11 in Dockerfile & Github actions

Iteron-dev:
After updating Django to 5.2, it is now possible to upgrade Python to 3.11.
Python 3.10 reaches end-of-life on 31 October 2026 (https://endoflife.date/python), so it's a good idea to upgrade sooner rather than later. Additionally, Python 3.11 offers significant performance improvements (https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-faster-cpython).

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -13,7 +13,7 @@ jobs:
     timeout-minutes: 60
 
     steps:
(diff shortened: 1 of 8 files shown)


twalen (approved):
Looks good.

PR #533, opened 2025-06-08
Optimize queries for participants_data view

### LINKED ISSUE #345: /participant_data/ too many queries
With 7 models referencing back to the Participant model and no prefetch_related()/select_related(), participants_data view produces a ton of queries, which is slow.

reivvax:
Closes #345


oioioi/participants/management/commands/import_participants_for_oi.py
twalen, line 1: please change name in such a way that it is clear that this is dev only script and it creates random data `dev_generate_random_oi_participants.py`

PR #556, opened 2025-09-27
Fix: Docker build failure (ModuleNotFoundError) and cypress tests

Iteron-dev:
For some strange reason, the [docker build action](https://github.com/sio2project/oioioi/blob/master/.github/workflows/build.yml) along with the [cypress tests action](https://github.com/sio2project/oioioi/blob/master/.github/workflows/cypress.yml) stopped working on Jun 18. A fix was made https://github.com/sio2project/oioioi/commit/50657ed17f02f255e21e8310de34bb415e400a98 - but it didn't help with Cypress, only with the docker build action. Unfortunately, as it turns out, the docker build action (including Cypress) still doesn't work due to another error:

```
0.280 Traceback (most recent call last):
0.280   File "/home/oioioi/.local/bin/oioioi-create-config", line 5, in <module>
0.280     from oioioi.deployment.create_config import main
0.280 ModuleNotFoundError: No module named 'oioioi'
```

This error probably didn't exist before, but now it does because of the installation of a newer version of `setuptools` through dependencies.

Long-term, it would be advisable to stop using `setup.py` in place of `pyproject.toml` and gently modernize the build process.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -58,22 +58,17 @@ RUN pip3 install --user bsddb3==6.2.7
 
 WORKDIR /sio2/oioioi
 
-COPY --chown=oioioi:oioioi setup.py requirements.txt ./
-RUN pip3 install -r requirements.txt --user filetracker[server]
-COPY --chown=oioioi:oioioi requirements_static.txt ./
-RUN pip3 install -r requirements_static.txt --user
+COPY --chown=oioioi:oioioi . ./
+RUN pip3 install --user -r requirements.txt filetracker[server]
+RUN pip3 install --user -r requirements_static.txt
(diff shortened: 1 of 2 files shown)


stopnoanime (commented):
Good fix

stopnoanime (dismissed):

Iteron-dev:
Since the Cypress tests were still failing ([https://github.com/Iteron-dev/oioioi/actions/runs/18063771807](https://github.com/Iteron-dev/oioioi/actions/runs/18063771807)), I removed the installation of the new `gevent==25.5.1` (honestly, I don’t know why such a new gevent ended up here), because the workers were throwing errors. I also fixed the insertion of `CAPTCHA_TEST_MODE=True` into `settings.py`.

Now it seems like everything is working as it should :)

PR #557, opened 2025-09-28
Add missing fixes from Ruff

Iteron-dev:
https://github.com/sio2project/oioioi/pull/549#discussion_r2204851336

Adding the missing fixes from Ruff along with removing the ignores that were added in https://github.com/sio2project/oioioi/pull/549. Some fixes required manual intervention. 
At the same time, part of `re_path` was replaced with `path`.
Ruff was also updated to version `0.13.2`.


twalen:
I will review it tomorrow (29.09)

Issue #564, opened 2025-10-06
Insufficient validation of field 'School year' in contest creation.


malevitzch:
<img width="907" height="959" alt="Image" src="https://github.com/user-attachments/assets/7bfcf512-9f60-401d-8405-00b175f6213b" />
How to reproduce: 
Create a contest inputting invalid data such as regular text into the 'School year' field, like shown below:
<img width="926" height="279" alt="Image" src="https://github.com/user-attachments/assets/dbb450a3-05b2-4624-bb09-6edf7227402b" />

PR #565, opened 2025-10-08
fixing label for school_year

twalen:
Partially solves issue #564


twalen (commented):
looks ok, but it would be nice to move the format into field hint

oioioi/contests/forms.py
```diff
@@ -46,7 +46,7 @@ def validate_years(year):
 
     school_year = forms.CharField(
         required=False,
-        label=_("School year"),
+        label=_("School year (format: yyyy/yyyy)"),
```
twalen, line 49: this requires new translation

PR #566, opened 2025-10-10
Docker optimizations

otargowski:
This PR decreases both the docker image size and its build time by:
- Not including the downloaded sandboxes twice in the final image by using an additional build stage.
- Purging texlive documentation, since Debian doesn't always put it in separate packages.
- Using ``no-install-recommends`` with `apt install` (which alone saves ~1GB), as up to now e.g. mupdf (a full-fledged pdf viewer) was included in the image. The only needed package installed this way  seems to be `tex-gyre`.
- Switching to `python:3.11-slim` as the base image. This required explicitly installing `make`, `g++` and `libc6-dev`.
- ~~Replacing `pip` with `uv pip`, which is almost a drop-in replacement and provides a 155 -> 8s speedup for the `requirements.txt` installation step on my modern cpu. On a i7 2640M from 2011 it takes 15s. The only issue was the lack of `--user` support, which was handled by using `.local/` as a venv.~~ --> this has been separated into #588.
- Parallelizing `upload_sandboxes_to_filetracker`, which is cpu-bound on the client size while compressing the already compressed sandboxes.
- Starting only filetracker for uploading the sandboxes to it, because `mailnotifyd` and `rankingsd` consume noticeable cpu while probably trying to access the non-existent DB.

The resulting development image size decrease is 7.08 -> 3.3 ~~3.14~~ GB.

The image could be shrunk further by removing e.g. old gcc sandboxes from the `Manifest` file, which might simply work since existing installations should already have all old sandboxes on filetracker. Alternatively, a separate `Manifest` file could be maintained.
Regardless, that is out of scope for this PR.

### DIFF
--- a/oioioi/sioworkers/management/commands/upload_sandboxes_to_filetracker.py
+++ b/oioioi/sioworkers/management/commands/upload_sandboxes_to_filetracker.py
@@ -1,9 +1,14 @@
-import os
-import os.path
+from concurrent.futures import ProcessPoolExecutor
+from pathlib import Path
 
 from django.core.management.base import BaseCommand
 
 from oioioi.filetracker.client import get_client
+
+
+def upload_sandbox(file):
+    filetracker = get_client()
+    filetracker.put_file("/sandboxes/" + file.name, str(file))
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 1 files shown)


otargowski:
I updated the PR to accomodate s3dedup. The uv migration will happen in another PR.
Tests fail on master too.

PR #593, opened 2025-12-02
replace external MathJax CDN with local offline-ready version (fixes #44)

### LINKED ISSUE #44: Stop using CDNs
Hello,
Could you please stop using external CDNs (exactly CloudFlare CDN to deliver `mathjax`) on SIO2?

kama-pa-go:
This merge integrates MathJax directly into the application using Webpack, removing the dependency on CloudFlare CDN. This ensures the platform is fully functional in offline deployment scenarios.

Key changes:
- Added `mathjax` (v2.7) to package.json and configured CopyPlugin in Webpack to serve assets locally from `dist_webpack`.
- Updated `settings.py` to point `MATHJAX_LOCATION` to local static files.
- Injected inline MathJax configuration in `base.html` to enable single-dollar (`$`) delimiters for inline math, which were previously unsupported by the default CDN configuration.
- Added `dist_webpack` to `STATICFILES_DIRS` to allow Django to discover the new assets.

Closes #44
Relate to #586

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "jquery": "^1.12.4",
     "js-cookie": "^3.0.5",
     "marked": "^15.0.7",
-    "mathjax": "^2.7.9",
+    "mathjax": "^4.0.0-beta.7",
     "select2": "^4.1.0-rc.0"
   }
 }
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -10,6 +10,7 @@
   "devDependencies": {
     "@eslint/js": "^9.23.0",
     "concurrently": "^9.1.2",
+    "copy-webpack-plugin": "^13.0.1",
```
twalen: it this library also needed during the deployment?

PR #599, opened 2025-12-15
Fixed the formatting of the limits in the contest tab

### LINKED ISSUE #560: Limits improperly displayed
Image

Time limit should be 10s and memory limit should be divided by 1024 and be 256MB

### LINKED ISSUE #600: Frontend problems with different time/memory limits for different languages
Likely due to the project's support for differentlimits for different languages for the same problem, the time/memory limits display is not well aligned inside the problems table
Image
This stems from the fact that, currently, the time + memory limits for each problem are a small, nested table inside the main problems table. The subtable has as many rows as there are different limits and is independent from the subtables, making the columns inside them misaligned as shown in the picture above. 
It is my understanding that the feature of having limit overrides for different programming languages is seldom used by problemsetters. 
If there are no overrides (when the creator simply just sets a time limit and a memory limit for the problem) then the subtables each have one row and the row contains three columns, one of which is empty.
It would likely be a reasonable solution to have the problems inside django contain a flag about whether or not they have "regular" or multi-row limits and then display the table accordingly, keeping the current behaviour for those who use multi-row limits and having a different one with separate time and memory limit columns which would provide proper alignment.

malevitzch:
The time limits are now elegantly formatted to avoid relying on floating point numbers. Memory limits are now displayed in mebibytes rather than megabytes (which seems to be what many problemsetters call megabytes).
image
Closes #560.
The alignment is off, but this is a largely unrelated problem (that had been present before) which stems from much deeper problems with how limits are handled, I am opening a separate issue for it.


twalen (approved):
Looks good

Issue #94, opened 2022-10-17
Admin page for Teachers - usability issues


twalen:
Admin view for managing teachers have several usability issues:
* missing search option (search by school or search by usename or first/last name)
* there is no timestamp of request date (so it is not possible to filter recent requests)
* adding new teacher form is not usable due to too large select field (static HTML element with all users)
* missing translations "Dodaj teacher" & "900 teachers"

Affected urls:
* https://szkopul.edu.pl/admin/teachers/teacher/
* https://szkopul.edu.pl/admin/teachers/teacher/add/

PR #602, opened 2025-12-17
Partial resolve for #94 -translation issues with admin/teachers

kama-pa-go:
Corrected translations:
* "Teacher" (and "Teachers")
* "Join date"
* "Active"
* "Save and add another"
* "School"

Relate to #586


oioioi/teachers/templates/teachers/change_form.html
```diff
@@ -0,0 +1,69 @@
+{% extends "admin/change_form.html" %}
+{% load i18n admin_urls %}
+
+{# Override buttons so they can be translated properly #}
```
twalen: is there any other way to inject context into the translations? right now this creates a code duplication and it is using unused buttons like archive / unarchive

PR #606, opened 2025-12-31
Feature: last submission status

m4teuk:
Added a badge to each problem row in the problems tab. The badge display the status of the users last submission for that problem, like "Initial tests: failed" or "Pending". It is displayed only for normal (not ignored submissions). 
image
image

### DIFF
--- a/oioioi/contests/views.py
+++ b/oioioi/contests/views.py
@@ -5,7 +5,7 @@ from django.conf import settings
 from django.contrib import messages
 from django.contrib.auth.models import User
 from django.core.exceptions import PermissionDenied, SuspiciousOperation
-from django.db.models import Q
+from django.db.models import OuterRef, Q, Subquery
 from django.http import HttpResponse, HttpResponseRedirect
 from django.shortcuts import get_object_or_404, redirect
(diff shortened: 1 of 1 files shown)


oioioi/contests/views.py
```diff
@@ -154,28 +154,51 @@ def problems_list_view(request):
...
+                user__id=request.user.id,
+                problem_instance_id__in=pi_ids,
+                kind="NORMAL" # ignore ignored submissions
+                )
+            .order_by("-date")
+        )
+        for s in submissions_qs:
+            pid = s.problem_instance_id
+            if pid not in last_submission_map:
+                last_submission_map[pid] = s
```
MasloMaslane: it shouldn't always fetch last submission, for example in some contests max submission counts as the score (OI, OIJ).

In this PR https://github.com/sio2project/oioioi/pull/405 there are functions implemented that return scored submission effectively. Maybe you can copy them here.

MasloMaslane (changes requested):
Also I don't really like that the badge is next to the title. Maybe it could be immediately to the left of the score?

MasloMaslane (changes requested):

PR #607, opened 2026-01-07
submit another testrun redirect

wlamandini:
Now the submit another button in a testrun submission redirects to the submit testrun page (instead of the normal submission page).


twalen (approved):
Looks good

Issue #587, opened 2025-11-19
Displaying partial scoring for problems with a checker


sylwia-sapkowska:
Currently, for a problem with partial scoring, status on a test is OK independently from percent of score achieved on a particular test. This may cause confusion for new users. I propose adding percent of scores next to the output from the checker below the test status table, e.g. 
Test 1a (30%): 1st line of output wrong, expected (...)

MasloMaslane:
As I wrote in https://github.com/sio2project/oioioi/issues/584, I also suggest changing the status text, because not all people scroll down to the comments

PR #611, opened 2026-01-14
Partial scoring enhancements

sylwia-sapkowska:
1. If the test doesn't pass fully (e.g. gets 50% points), the message in final testing report is: "Partially OK." (partially solves #587).
2. The comments from the checker were extended in case of exceeding half the time limit and displayed when status is different than TLE.

### DIFF
--- a/oioioi/programs/tests.py
+++ b/oioioi/programs/tests.py
@@ -3,7 +3,6 @@ import re
 import urllib
 from collections import defaultdict
 from datetime import UTC, datetime, timedelta  # pylint: disable=E0611
-from unittest import mock
 
 import pytest
 import six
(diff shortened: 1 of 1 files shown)


oioioi/programs/models.py
```diff
@@ -312,6 +312,45 @@ class TestReport(models.Model):
...
+    def get_comment_display(self):
+        print(self.comment, self.time_used, self.test_time_limit, self.should_display_comment())
+        comment = self.comment if self.comment else ""
+        if comment and not comment.endswith("."):
+            comment += "."
+        if self.status != "TLE":
+            try:
+                # In this case threshold linear scoring was used.
+                if self.score < self.max_score and self.time_used * 2 > self.test_time_limit:
+                    comment += _(" Half of the time limit exceeded.")
```
MasloMaslane: What about contests other that OI? Were linear threshold is not used

oioioi/programs/models.py
```diff
@@ -312,6 +312,45 @@ class TestReport(models.Model):
...
+        if self.status != "TLE":
+            try:
+                if self.score < self.max_score and self.time_used * 2 > self.test_time_limit:
+                    return True
+            except (TypeError, AttributeError):
+                pass
+        return False
+
+    def get_comment_display(self):
+        print(self.comment, self.time_used, self.test_time_limit, self.should_display_comment())
```
MasloMaslane: Please remove debug

PR #615, opened 2026-01-16
Optimize getting language versions of newsfeeds

otargowski:
Now there are 2 queries executed for fetching newsfeed items instead of O(items), as `prefetch_related` actually works now.
This seems to have cut down the load on SZKOpuł by a few units (most of the requests are `GET /` for some reason).


twalen (approved):
Looks good

PR #622, opened 2026-02-10
Fixed the build failing due to coreapi relying on setuptools features removed in 82.0.0

### LINKED ISSUE #621: ModuleNotFoundError: No module named 'pkg_resources' when building the project
When running the automated unit tests for #599, the build failed in an unexpected manner. As far as I am aware, the PR does not modify anything that could potentially cause a build failure of this type 

I tried to locally build the branch and it failed with the same error.
I also tried to build the master branch and it failed as well, so this seems like an urgent issue.
Image

I attach the logs of the initial unit test run below
[logs_56941380660.zip](https://github.com/user-attachments/files/25205492/logs_56941380660.zip)

malevitzch:
The (temporary) solution is to pin the setuptools version to be <81 as recommended by setuptools itself.

The long-term solution is to migrate from CoreAPI to OpenAPI. According to the CoreAPI github page:
image

Closes #621.

I will create an issue and try to perform the migration later since the project's usage of CoreAPI is very limited and I am hoping that it shouldn't be too difficult.

### DIFF
--- a/.github/actions/setup-env/action.yml
+++ b/.github/actions/setup-env/action.yml
@@ -35,4 +35,5 @@ runs:
   - name: Install python dependencies
     run: |
       uv pip install -r requirements.txt
+      uv pip install "setuptools<81"
     shell: sh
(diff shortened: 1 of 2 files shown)


twalen (approved):
Looks good as a workaround.

Issue #574, opened 2025-11-01
Show memory usage in submission reports


otargowski:
Per-test memory usage should be shown in submission reports. It is reported by sioworkers in the 'mem_used' env field just like 'time_used', but it isn't saved anywhere by oioioi.

The usage should probably be rounded to MiB, maybe if the limit is below 64MiB, then to KiB. The limit must also be rounded, since many packages don't have multiples of 1024 as the limit and we don't want "24 / 501.1847288528 MiB" like in testruns (has this been fixed?).
I think rounding down will be best.

If the submission is old and therefore didn't have the mem usage saved for any test, then the report should look the same as now.

Requested by @tonowak.

Good first issue for TAG.

PR #660, opened 2026-03-20
Show memory usage

wlamandini:
Solves issue https://github.com/sio2project/oioioi/issues/574.

The "Memory" column in the submission report looks like this:
Screenshot from 2026-03-20 19-26-05
Screenshot from 2026-03-20 19-38-01
Screenshot from 2026-03-20 19-38-21

I would be thankful for some feedback regarding the format of the displayed memory usage and memory limit. I have chosen the units to always be MiB, since the memory limit should be of this order, and the user's code isn't able to achieve less than ~0.9MiB memory used, unless the program crashed immediately with a signal (e.g. big global array), then 0MiB is recorded. I show 1 decimal place if the number of MiBs is less than 10, otherwise only the integer part.

I show the "Memory" column if at least 1 test has nonzero memory usage. This condition will be false if the memory usage was not measured. However, this has a funny consequence, that if the user's code crashes immediately on all tests, then the column also isn't shown.


MasloMaslane (approved):

twalen (approved):
Looks good.

tonowak:
thanks!

PR #676, opened 2026-04-04
fix: hide user name in forum votes for anonymous contest registrations

matthberendt:
A bug fix for the OIOIOI forum
Reaction lists (the "reacted by" section on posts) were showing participants' full real names even when they'd registered anonymously for a contest. The fix hooks the template tag into the contest's existing anonymity logic, so anonymous participants now only appear by username. If there's no contest context at all, it falls back to the old behavior as a safety net.


malevitzch:
I'll try to fix it up so ruff stops complaining and review it tomorrow

matthberendt (commented):
Looks good.

twalen (approved):
Looks good

PR #680, opened 2026-04-08
subtask dependency

### LINKED ISSUE #591: Add subtasks dependency
In some problems, often subtasks depend on each other,

e.g.  in the following case

1 | A, B <= 20 |
2 | A <= 20 |
3 | A <= 500 | 
4 | B <= 20 |
5 | No additional constraints. |

2 depends on 1, 3 depends on 1 and 2, 4 depends on 1, and 5 depends on all the others.

It should be possible to integrate this in a way that if there is a failing test in group 1, groups 2, 3, 4, 5 also fail, without needing to add the same test in all groups. Note that we don't want to actually run these tests again, just append their scores to the proper groups.

This probably should be added in `config.yml` of the problem, for instance.

m4teuk:
- Subtask dependency config option
- Displaying a warning icon and an explanation when the score is lowered due to a failing prerequisite subtask


oioioi/_locale/locale/pl/LC_MESSAGES/django.po
```diff
@@ -11076,3 +11076,8 @@ msgstr "Liczba węzłów"
 #: zeus/templates/zeus/program-report.html:21
 msgid "Zeus id"
 msgstr "ID Zeusa"
+
```
twalen: please update translations via transifex

sylwia-sapkowska:
sinol-make docs should be updated accordingly, to describe how to use it.

MasloMaslane:
Not only the docs, but also scoring and config validation should be updated. Would be great if someone would be willing to do this:)

oioioi/programs/handlers.py
```diff
@@ -406,6 +406,98 @@ def grade_groups(env, **kwargs):
...
+        combined_results = dict(own_tests)
+        for prereq in prereqs:
+            for prereq_test_name, prereq_result in group_test_results.get(prereq, {}).items():
+                prereq_max = ScoreValue.deserialize(prereq_result["max_score"])
+                prereq_score = ScoreValue.deserialize(prereq_result["score"])
+                if prereq_max is None or prereq_max.value == 0:
+                    normalized_value = 0
+                else:
+                    # Scale prereq score proportionally to the target max_score
+                    normalized_value = round(prereq_score.value * target_max_value / prereq_max.value) if prereq_score else 0
```
sylwia-sapkowska: Ensure there are no rounding issues here, as errors can accumulate across multiple tests. Instead of scaling by the maximum score within this group, consider rescaling using the least common multiple of all group maxima (though this may become very large), or use exact percentage values directly from TestReport to avoid loss of precision (for problems with partial scoring available)

oioioi/sinolpack/package.py
```diff
@@ -1110,6 +1111,74 @@ def _assign_scores(self, scored_groups, total_score_if_auto):
...
+                            "referenced by group '%(group)s' is not a valid scored group. "
+                            "Valid groups: %(groups)s"
+                        ) % {"prereq": prereq, "group": group, "groups": sorted(scored_groups)}
+                    )
+
+        # Detect cycles using DFS
+        visiting = set()
+        visited = set()
+
+        def dfs(node):
```
sylwia-sapkowska, line 1158: Looks good!

sylwia-sapkowska (changes requested):

sylwia-sapkowska (approved):
Looks good to me.

Issue #683, opened 2026-04-08
Reupload should fail when task package type differs from original task


wlamandini:
Currently it is possible to:
- reupload an interactive task package for a non-interactive task,
- reupload a normal task package for a quiz task, by manually entering a problem instance number in the url for reuploading.

Both actions above succeed without any errors. However, the task is still judged in the original way (non-interactively in the first case, and quiz-like in the second case), even though the problem's files are from the new package (statement, tests, interactive checker).

The expected behavior for a reupload like this is to simply fail with an appropriate error, since handling changes in the problem type seems unnecessary.

PR #685, opened 2026-04-08
Fix: Reuploading a task package of a different type than the existing task

matthberendt:
Fix regarding this bug: https://github.com/sio2project/oioioi/issues/683

1. Added Task Type Validation: In oioioi/sinolpack/package.py, when reuploading to an existing_problem, the system now calls _ensure_compatible_controller. This method compares the existing problem's controller with the new package's determined controller. If they do not match, it blocks the reupload by raising a ProblemPackageError.

2. Improved Error Reporting:_create_problem_or_reuse_if_exists, _ensure_short_name_equality_with, and the new _ensure_compatible_controller with @_describe_processing_error. This ensures that if the task types mismatch these changes effectively close the gaps allowing interactive packages to overwrite non-interactive tasks and normal packages to overwrite quiz tasks, ensuring task integrity during reuploads.

### DIFF
--- a/oioioi/sinolpack/package.py
+++ b/oioioi/sinolpack/package.py
@@ -16,6 +16,7 @@ from django.contrib.auth.models import User
 from django.core.exceptions import ValidationError
 from django.core.files import File
 from django.core.validators import slug_re
+from django.utils.module_loading import import_string
 from django.utils.translation import gettext as _
 
 from oioioi.base.utils import generate_key, naturalsort_key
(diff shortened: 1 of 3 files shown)


matthberendt:
Ruff formatted

oioioi/sinolpack/package.py
```diff
@@ -332,13 +336,35 @@ def _create_problem_or_reuse_if_exists(self, existing_problem):
...
                 _("Tried to replace problem '%(oldname)s' with '%(newname)s'. For safety, changing problem short name is not possible.")
                 % {"oldname": existing_problem.short_name, "newname": self.short_name}
             )
 
+    @_describe_processing_error
+    def _ensure_compatible_controller(self, existing_problem):
+        """Checking if task type didn't change."""
+        new_controller_name = self._get_controller_name()
+        if existing_problem.controller_name != new_controller_name:
+            try:
```
twalen, line 353: is it possible to write this without try/except (like with hasattr)

twalen (changes requested):
could you add some regression tests?

twalen:
+ please resolve merge conflict

PR #697, opened 2026-04-20
added save success message

### LINKED ISSUE #654: Contest Administration / Settings - No Save "success" message
When in the Settings tab inside the Contest administration directory, clicking the save button after making some changes only refreshes the page. No "success" message.

Maybe it would be nice to display a banner saying something like "settings saved successfully".

m4teuk:


oioioi/contests/admin.py
```diff
@@ -264,6 +266,11 @@ def response_change(self, request, obj):
         # Never redirect to the list of contests. Just re-display the edit
         # view.
         if "_popup" not in request.POST:
+            msg = _('The %(name)s "%(obj)s" was changed successfully.') % {
+                "name": capfirst(self.opts.verbose_name),
+                "obj": str(obj),
```
twalen, line 271: passing str(obj) might be give weird messages to a user, maybe it is enough to return generic message (even without model name), like 'The object was changed successfully'?

otargowski: The more verbose behaviour is the (django?) default and is present in other admin interfaces.

otargowski (approved):
LGTM 👍👍👍

twalen (approved):

PR #699, opened 2026-04-20
Add option of searching tags by their abbreviated name

### LINKED ISSUE #681: Add option of searching tags by their abbreviated name
It should be possible to search for tags using their short/abbreviated/commonly used name (like "binsearch", "deque"), not only by their official names (like "Wyszukiwanie binarne po wyniku", "Kolejka dwustronna").

sylwia-sapkowska:
Solves #681. Adds option to search tag by their commonly used (short) name.

tag-improvement

### DIFF
--- a/oioioi/problems/tests/test_tags.py
+++ b/oioioi/problems/tests/test_tags.py
@@ -1,9 +1,11 @@
+import json
 import urllib.parse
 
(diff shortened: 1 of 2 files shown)


dec0dedd (approved):
Looks good to me

PR #706, opened 2026-04-21
Reupload task type check

wlamandini:
Solves issue https://github.com/sio2project/oioioi/issues/683.
Added task type change detection during unpacking of a reuploaded package.


oioioi/problems/problem_site.py
```diff
@@ -241,6 +241,7 @@ def problem_site_settings(request, problem):
             "model_solutions": model_solutions,
             "can_admin_problem": can_admin_problem(request, problem),
             "extra_actions": extra_actions,
+            "unsupported_reuploads": problem.controller_name == "oioioi.quizzes.controllers.QuizProblemController",
```
twalen: this should be a function in the controller

oioioi/sinolpack/package.py
```diff
@@ -339,6 +340,16 @@ def _ensure_short_name_equality_with(self, existing_problem):
                 % {"oldname": existing_problem.short_name, "newname": self.short_name}
             )
 
+    def _ensure_task_type_equality_with(self, existing_problem):
+        existing_problem_task_type = (
+            TaskType.INTERACTIVE if existing_problem.controller_name == "oioioi.sinolpack.controllers.SinolInteractiveProblemController" else TaskType.STANDARD
```
twalen: retrieving a task type could be a function in the controller?

oioioi/problems/problem_site.py
wlamandini: done

oioioi/sinolpack/package.py
wlamandini: done

twalen (approved):

twalen (approved):

PR #707, opened 2026-04-22
Bootstrap update

malevitzch:
Update to bootstrap 5


malevitzch:
Bootstrap 5 seems to work fine, but updating to jquery 4 was not possible, likely due to bootstrap-3-typeahead. Upon inspection, the code for typehaead (especially in tag_selection.js) is horrifying cursed mess and there doesn't seem to be a good drop-in replacement for how bootstrap-3-typeahead is used there.

twalen (approved):
Looks good

PR #716, opened 2026-04-29
Rejudge UI fix

### LINKED ISSUE #705: Partial rejudge is not partial without previewing count
Context: In https://github.com/sio2project/oioioi/pull/644 we added date and last submission filtering to the rejudge screen.
With it, a very nice "Preview count" button was added (screen below).

If that button is used, then everything works fine.
However, if the count isn't previewed, all submissions are rejudged, as if the filters were never applied.

This is misleading, and clearly unintended behaviour.

Jankwi:
This directly addresses https://github.com/sio2project/oioioi/issues/705, by declaring both parameters once for both previewing count and confirming the rejudge.

Closes #705


malevitzch (approved):
Looks good

PR #727, opened 2026-05-13
explaining lower scores

m4teuk:


oioioi/programs/utils.py
```diff
@@ -7,6 +7,7 @@
 from django.core.exceptions import PermissionDenied
 from django.shortcuts import get_object_or_404
 from django.urls import reverse
+from django.utils.translation import gettext
```
twalen: `from django.utils.translation import gettext as _`

dec0dedd (approved):
Looks good to me

PR #730, opened 2026-05-13
Add example tests endpoints

Arapak:
Two new endpoints:
`GET .../problem/<problem_id>/example_tests` - returns a json array with test names and urls
`GET .../problem/<problem_id>/<test_filename>` - streams the example test file

### DIFF
--- a/oioioi/problems/urls.py
+++ b/oioioi/problems/urls.py
@@ -22,6 +22,16 @@ problem_site_patterns = [
         views.problem_site_external_attachment_view,
         name="problem_site_external_attachment",
     ),
+    path(
+        "example_tests/",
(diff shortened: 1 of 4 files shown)


oioioi/problems/urls.py
```diff
@@ -22,6 +22,16 @@
         views.problem_site_external_attachment_view,
         name="problem_site_external_attachment",
     ),
+    path(
```
twalen, line 25: I prefer to handle such requests via API (api.py)

dec0dedd (dismissed):
Looks good to me

twalen (changes requested):

PR #735, opened 2026-05-17
Removed oioioi/cypress_settings.py

### LINKED ISSUE #559: Unused `cypress_settings.py` file
I noticed that the file `oioioi/cypress_settings.py` appears to be unused. At the moment, Cypress test settings are applied via `./easy-toolbox cypress-apply-settings`

It would make sense to update the workflow so that `cypress_settings.py` is actually used during Cypress testing. This could simplify configuring the test environment.

If that’s not feasible, then I suggest removing the file to avoid confusion.

dec0dedd:
Closes #559


m4teuk (approved):
Looks okay

twalen (approved):

PR #740, opened 2026-05-20
Add non-negative submissions limit validator

wlamandini:
Makes entering negative submissions limits no longer possible.


MasloMaslane (dismissed):

oioioi/contests/models.py
```diff
@@ -406,6 +406,7 @@ class ProblemInstance(models.Model):
         default=settings.DEFAULT_SUBMISSIONS_LIMIT,
         help_text=_("Use 0 for unlimited submissions."),
         verbose_name=_("submissions limit"),
+        validators=[MinValueValidator(0, "Submissions limit must be a non-negative number.")],
```
twalen, line 409: please add regression test

twalen: also please add a migration

twalen (changes requested):

oioioi/contests/models.py
wlamandini: added test and migration

twalen (approved):

PR #756, opened 2026-06-03
More performance optimizations

otargowski:
A continuation of #690. A few last views were improved to use O(1) queries, most notably the submit view (I found a not too invasive way with `prefetch_related`) and contest rules view.

Other than that, many `prefetch_related` calls were replaced with new `annotate_known_related[_many]` utilities, which make use of already known related objects, e.g. for `Submission.objects.filter(problem_instance__in=problem_instances)`, we can "annotate" `problem_instances`. This has the huge benefit of not needing to prefetch transitively related objects all the time, like `round`, `problem` and `contest` in the example, which makes the code cleaner and less prone to someone forgetting to prefetch some relation in the future.

`eval_contest_submissions_qs_with_common_related` was also introduced, which annotates submissions with already fetched problem instances with all their related objects and also prefetches stuff for `oioioi.scoresreveal` if it is enabled (now this ugly `if` is in one place instead of a few).

There are also a few other small optimizations and cleanups.


oioioi/portals/views.py
```diff
@@ -323,7 +323,8 @@ def delete_portal_view(request):
 @request_cached
 def my_portal_url(request):
     try:
-        portal = Portal.objects.select_related("root", "owner").get(owner=request.user)
+        portal = Portal.objects.select_related("root").get(owner=request.user)
+        portal.owner = request.user
```
otargowski, line 328: This trick for avoiding a select_related could use a comment...

oioioi/programs/controllers.py
```diff
@@ -845,8 +839,8 @@ def render_submission_footer(self, request, submission):
         )
 
     def get_allowed_languages_for_problem(self, problem):
-        # This query is a bottleneck for the submit view, see comment for `._add_langs_to_form()`.
-        allowed_langs = list(ProblemAllowedLanguage.objects.filter(problem=problem).values_list("language", flat=True))
+        # This should be fetched in bulk with prefetch_related.
```
otargowski: This comment could be better, like the one for `_get_problem_compilers_cached`.

oioioi/contests/views.py
```diff
@@ -579,15 +567,17 @@ def contest_files_view(request):
...
     if is_admin:
         problem_ids_without_admin = {pi.problem_id for pi in visible_problem_instances(request, no_admin=True)}
     else:
         problem_ids_without_admin = set(problem_ids)
-    problem_files = ProblemAttachment.objects.filter(problem_id__in=problem_ids).select_related("problem").prefetch_related("problem__names")
+    problem_files = ProblemAttachment.objects.filter(problem_id__in=problem_ids)
+    problem_files = annotate_known_related_many(problem_files, "problem", problems)
 
-    round_file_exists = contest_files.filter(round__isnull=False).exists()
-    add_category_field = round_file_exists or problem_files.exists()
```
otargowski, line 622: The `round__isnull=False` condition was forgotten.

oioioi/base/utils/__init__.py
```diff
@@ -644,3 +648,50 @@ def union(elem1, elem2):
...
+
+
+def annotate_known_related(objs, related_name, related):
+    objs = list(objs)
+    for obj in objs:
+        setattr(obj, related_name, related)
+    return objs
+
+
+def annotate_known_related_many(objs, related_name, related_list):
```
otargowski: These functions could use a comment explaining the purpose and possible uses.

oioioi/contests/utils.py
```diff
@@ -128,6 +129,16 @@ def get_key_for_comparison(self):
         )
 
 
+def rounds_in_contest(contest):
```
otargowski, line 135: There should be a comment explaining why the argument is contest instead of request (which would make caching trivial), as it is needed for methods which don't have a (proper) request, that are used by e.g. rankings stuff.

otargowski (commented):
I found a few small flaws, which will be fixed soom, as will the merge conflict.

PR #771, opened 2026-06-21
Bootstrap admin panel tabular fix

malevitzch:
Bootstrap 5 made the 'row' class add negative margins which caused certain parts of the UI to be almost unusable. Removing row seems to fix the issue without breaking anything.

Before and after:
image

image


malevitzch:
@otargowski I think this was one of the things annoying you, removing the 'row' in tabular.html makes the UI usable

otargowski:
Looks great now! `oioioi/base/templates/nesting/admin/inlines/tabular.html` should probably be fixed in a similiar fashion.

malevitzch:
Ruff randomly failed for me here with error code 404 so I had to rerun it... not sure if it's an issue on our part, it just failed to access the action so it's probably on astral's side.

Anyways, the other fix is now applied too

twalen (approved):

PR #775, opened 2026-06-26
More frontend fixes

malevitzch:
Fixed the spacing between the search bar and the magnifying glass icon.


malevitzch:
I wanted to fix the bug with Task Archive too, but unfortunately I cannot properly reproduce it on my machine. As far as I am concerned, that is the last bootstrap-related problem

twalen (approved):

PR #777, opened 2026-07-02
Add personal data confirmation for OI final participants

mmathiasT:


oioioi/_locale/locale/pl/LC_MESSAGES/django.mo
twalen, line 1: please remove all changes from _locale directory, this should be handled by separate PR (via transifex)

oioioi/default_settings.py
```diff
@@ -241,6 +241,7 @@
     'django.contrib.messages.middleware.MessageMiddleware',
     'dj_pagination.middleware.PaginationMiddleware',
     'oioioi.contests.middleware.CurrentContestMiddleware',
+    'oioioi.oi.middleware.OIDataConfirmationMiddleware',
```
twalen: do we need that for all instances of oioioi?

PR #786, opened 2026-08-22
Fix random failures when creating tag proposals in mass_create_tool

### LINKED ISSUE #649: Nondeterministic mass_create_tool test
The test "test_long_flags" for mass_create_tool failed randomly on an unrelated commit due to extreme misfortune and the generator failing to get an unique candidate 10 times in a row. This is really dumb, something should probably be done about this since it's not a very good test if something like that can happen (the test passed before merge and then failed after)

https://github.com/sio2project/oioioi/actions/runs/22960718126/job/66650241758
Image
Image

Staheos:
Fixes #649.

`mass_create_tool` retried random proposal candidates and failed after ten collisions, even when unused combinations were still available. This made proposal generation fail randomly as the requested count approached the number of possible combinations.

Draw unique combination indices without replacement and reject counts that exceed the available space. Add regression tests that fill the entire proposal space and verify invalid counts.

### DIFF
--- a/oioioi/problems/management/commands/mass_create_tool.py
+++ b/oioioi/problems/management/commands/mass_create_tool.py
@@ -390,34 +390,43 @@ class Command(BaseCommand):
 
     def create_proposals(self, count, problems, users, tags, proposal_model, verbose_name, verbosity):
         """
-        Creates algorithm tag proposals by pairing problems, users, and algotags randomly.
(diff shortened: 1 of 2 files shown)


malevitzch:
Something needs to finally be done about SonarCloud analysis failing due to "weak cryptography" on tests, lmao

twalen (approved):

PR #787, opened 2026-08-24
Speed up user group admin form

### LINKED ISSUE #711: Usergroup add view lists all users and so is slow
Example URL is `https://szkopul.edu.pl/admin/usergroups/usergroup/add/`.
It renders a somewhat fancy `select` field filled with ALL users and loads very slow.
It should use something like the `user_hints` stuff.

Staheos:
Fixes #711.

The user group admin form used `filter_horizontal`, which rendered all member candidates and all eligible owners when the page was opened. This switches both fields to Django autocomplete widgets.

Owner suggestions remain limited to teachers and superusers. The autocomplete endpoint uses `OioioiUserAdmin`, so the same restriction is applied there as well.

Tests:
- `pytest oioioi/usergroups/tests.py`
- `ruff check oioioi/usergroups/admin.py oioioi/usergroups/tests.py`
- `ruff format --check oioioi/usergroups/admin.py oioioi/usergroups/tests.py`

### DIFF
--- a/oioioi/usergroups/admin.py
+++ b/oioioi/usergroups/admin.py
@@ -18,7 +18,7 @@ def get_user_name_and_login_bounded(self, user):
 
 class UserGroupAdmin(admin.ModelAdmin):
     exclude = ("addition_config", "sharing_config", "contests")
-    filter_horizontal = ("owners", "members")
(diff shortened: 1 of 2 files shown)


otargowski (approved):
Looks good! 🚀

PR #788, opened 2026-08-25
Fixed timelines being messy after update to bootstrap 5

### LINKED ISSUE #774: Timelines are ugly and unusuable
When navigating to "Contest Administration" -> "Timeline", you can see something like this:
Image

Not sure what this is caused by, but it looks horrible and something should probably be done about it. I don't think the timeline is used too often, so it could perhaps be moved to oioioi-contrib

malevitzch:
There were some bootstrap problems after all, notably that bootstrap 5 expects input fields to be direct children of the input group rather than wrapped in a div.

Before:
image
After:
image

Closes #774


otargowski:
Looks good except for the calendar icon turning white (so almost invisible) on hover.

malevitzch:
Should be better now, unhovered and hovered for reference
image
image

otargowski (approved):

PR #789, opened 2026-08-29
Fix signal hints in test runs

### LINKED ISSUE #354: Submission error hints are not appearing for testruns

Staheos:
Fixes #354.

Signal hints were only extracted from regular test reports and only recognized the `process exited due to signal` format. Test runs use a separate report template, while sioworkers can also produce `program exited due to signal`.

This recognizes both formats in one place and passes the extracted signal to the test run template, reusing the existing error hint.

The regression tests fail without the implementation and pass with it.

Tests:
- `pytest oioioi/programs/tests.py::TestReportDisplayTypes::test_signal_hint oioioi/testrun/tests.py::TestTestrunViews::test_signal_hints`
- `pytest oioioi/programs/tests.py oioioi/testrun/tests.py`
- `ruff check oioioi/programs/controllers.py oioioi/programs/tests.py oioioi/testrun/controllers.py oioioi/testrun/tests.py`
- `ruff format --check oioioi/programs/controllers.py oioioi/programs/tests.py oioioi/testrun/controllers.py oioioi/testrun/tests.py`

### DIFF
--- a/oioioi/programs/controllers.py
+++ b/oioioi/programs/controllers.py
@@ -101,6 +101,23 @@ def get_report_display_type(request, test_report):
     return display_type
 
 
+def get_signal_from_comment(comment):
+    # Sioworkers doesn't give us exit codes or signals explicitly. Neither does sio2jail.
+    # This detection mechanism is similar to the one sioworkers uses to give a RE verdict:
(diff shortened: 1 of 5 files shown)


otargowski (approved):
Looks great!

PR #790, opened 2026-09-13
Show cpu_exec load in worker load chart

### LINKED ISSUE #767: Show cpu_exec load in worker load chart
Currently, the live-updating charts in the "Manage workers" admin view show only capacity and load, where load interprets the `is_running_cpu_exec` worker info key as full load of a worker, which is reasonable (a cpu_exec / real-time task execution blocks the execution of other tasks on the given worker).

If the load resulting from real-time tasks was also shown on the charts, then they could be easier to interpret properly.

Perhaps a non-solid pattern could be used for this value instead of a color different from the one for load.

This is blocked by #757.

Staheos:
This separates the CPU-exec part of worker load in the load chart. The existing total load remains unchanged, CPU-exec load is returned separately and displayed with a hatched fill.

The regression test covers a regular worker running one task alongside a CPU-exec worker that consumes all of its concurrency.

Fixes #767

Screenshot:
download


otargowski (approved):
Looks great, I like the new chart 👍.

twalen (approved):

